PR #18, opened 2024-03-08
Ahmad kashkoush/i9/format text method

ahmad-kashkoush:
# Changes
- add format to English text as option for number types in drafting
# closes 
https://github.com/accordproject/template-playground/issues/9


sanketshevkar:
Can add some test cases for the changes you have made?

dselman:
I think the first step should be evaluate the existing open source libraries that do this. E.g.
https://www.npmjs.com/package/to-words

We want to find a high-quality, maintained, library, with a permissive license and with good locale support.

Once we've selected the library we would like to use then we can move onto to integrating it with drafters, dealing with configuration etc.

Yes, test cases will be very important, as will documentation.

ahmad-kashkoush:
so , I should be import https://www.npmjs.com/package/to-words and use it on format options, is that right?
and the test cases I will add a single file in the **test** , is that right also?
@dselman @sanketshevkar

ahmad-kashkoush:
I'm working on importing the following https://www.npmjs.com/package/to-words library, I'm wondering how can I get Locale without adding a parameter in drafter format Method?

```js
export function draftIntegerFormat(value:number,format:string, /* I need to add parameter for locale*/) : string {
    if(format === 'word'){
        const converter: ToWords = new ToWords();
        const res=converter.convert(value, /*format options*/);
        return res;
    } else{
 ```

**Comment by sanketshevkar** [comment]
> I'm working on importing the following https://www.npmjs.com/package/to-words library, I'm wondering how can I get Locale without adding a parameter in drafter format Method?

Why do you not want to add a new parameter?

**Comment by ahmad-kashkoush** [comment]
> > I'm working on importing the following https://www.npmjs.com/package/to-words library, I'm wondering how can I get Locale without adding a parameter in drafter format Method?
> 
> Why do you not want to add a new parameter?

to specify locale that I'll format based on

**Comment by sanketshevkar** [comment]
> to specify locale that I'll format based on

The question why would NOT want to add? I think you should be fine adding a new parameter, you'd just might want to fix the breaking test cases.

**Comment by sanketshevkar** [comment]
@dselman can you please review this PR?

File: src/drafting/Double/format.ts [hunk 1]
```diff
@@ -32,19 +32,25 @@ export function draftDoubleIEEE(value:number) : string {
  * @returns {object} the parser
```
sanketshevkar, line 32:
```suggestion
 * @returns {string} formatted double value as string
```

src/drafting/Integer/format.ts
```diff
@@ -28,14 +30,20 @@ export function draftInteger(value:number) : string {
  * @returns {object} the parser
```
sanketshevkar, line 30:
```suggestion
 * @returns {string} formatted integer value as string
```

sanketshevkar (changes requested):

sanketshevkar:
Added suggestion, I know these are not coming from your code changes, but might as well update these.
Thanks!

ahmad-kashkoush:
> > to specify locale that I'll format based on
> 
> The question was, why would NOT want to add? I think you should be fine adding a new parameter, you'd just might want to fix the breaking test cases.

sorry for misunderstanding :laughing: ,  I was a little bit skeptical about making changes at the method signature which may require lots of changing of the way we call the function on different files.

sanketshevkar:
> sorry for misunderstanding 😆 , I was a little bit skeptical about making changes at the method signature which may require lots of changing of the way we call the function on different files.
but ok, thanks

It okay 😄. Don't worry I can relate 😅

sanketshevkar:
Also please fix the DCO signature for this PR.
https://github.com/accordproject/template-engine/pull/18/checks?check_run_id=22605601318

src/drafting/Double/format.ts
```diff
@@ -29,22 +29,28 @@ export function draftDoubleIEEE(value:number) : string {
...
-            res += sep2 + d;
-        }
-        let i = vs.substring(0,vs.length - (len === 0 ? 0 : len+1));
-        while (i.length > 3) {
-            res = sep1 + i.substring(i.length - 3) + res;
-            i = i.substring(0, i.length - 3);
-        }
-        return i + res;
-    });
+    if(format==='word') {
```
dselman, line 35: Let's create an enumeration for this, and the values can be:

enum DraftFormat {
   NUMBER,
   TEXT
}

You can set the default value for the `format` to be `DraftFormat.NUMBER`.

dselman (changes requested):
Looking good. Just a small change for developer experience.

src/drafting/Double/format.ts
ahmad-kashkoush: thanks for reviewing, I've applied changes, but added default values for the enum, because **Number format** needs to use string method (replace).

ahmad-kashkoush:
Hi @dselman , I've created a more organized and cleaner new PR , 

https://github.com/accordproject/template-engine/pull/22

PR #21, opened 2024-04-15
feat: use child process to eval JS

dselman:
Adds a new generation option to the engine: `childProcessJavaScriptEvaluation` which will fork a new child Node.js process and eval the code within this new process. This ensures that user code (condition nodes, clause nodes, formulae) cannot access memory/crash/DDoS the host process.

Because JS eval is now async, some other refactoring had to be performed on the code that transforms the templatemark JSON tree to an agreementmark tree — gathering all the JS code and evaluating it, before starting the tree -> tree transformation.

### DIFF
--- a/src/JavaScriptEvaluator.ts
+++ b/src/JavaScriptEvaluator.ts
@@ -38,7 +38,7 @@ type WorkItem = {
 export type EvalResponse = {
     result: any // success if not null
     timeout?: boolean // if true the promise is rejected due to timeout
-    starvaton?: boolean // if true the promise is rejected due to lack of child process
+    starvation?: boolean // if true the promise is rejected due to lack of child process
     message?: string; // if promise rejected due to a caught exception this will be set
     elapsed?: number; // the elapsed time in ms to process the work item
     maxQueueDepthExceeded?: boolean // if true the promise is rejected because the queue is full
@@ -158,10 +158,10 @@ export class JavaScriptEvaluator {
             }
             // on timeout will send SIGTERM
             const worker = child_process.fork('./dist/worker.js', { timeout: options.timeout, env: {} });
-            this.workers.push(worker);
             if (!worker.pid) {
                 throw new Error('Failed to fork child process');
             }
+            this.workers.push(worker);
             work.pid = worker.pid;
             let result: any;
             worker.on('error', (err: any) => {
--- a/src/JavaScriptEvaluator.ts
+++ b/src/JavaScriptEvaluator.ts
@@ -106,7 +106,7 @@ export class JavaScriptEvaluator {
      * @param {EvalOptions} options the options for the request
      * @returns {Promise<EvalResponse>} the async result
      */
-    evalChildProcess(request: EvalRequest, options: EvalOptions = { timeout: 10000 }): Promise<EvalResponse> {
+    async evalChildProcess(request: EvalRequest, options: EvalOptions = { timeout: 10000 }): Promise<EvalResponse> {
         // console.log('Workers count: ' + this.workers.length);
         return new Promise((resolve, reject) => {
             const now = new Date().getTime();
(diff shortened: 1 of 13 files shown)


src/worker.js
```diff
@@ -0,0 +1,44 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+import dayjs from 'dayjs';
+import jp from 'jsonpath';
+
+/**
+ * This executed JS code inside a child process, so that the code
+ * has no access to the parent process and can be terminated on timeout etc
```
ekarademir, line 19: It doesn't have access to parent process, but it would have the same privileges, no?

ekarademir (approved):

mttrbrts (approved):

src/TemplateMarkInterpreter.ts
```diff
@@ -67,29 +77,31 @@ function evaluateJavaScript(clauseLibrary:object, data: TemplateData, fn: string
...
     const expression = fn.substring(fn.indexOf('{') + 1, fn.lastIndexOf('}'));
     if(expression.trim().length === 0) {
         throw new Error('Empty expression');
     }
     try {
-        const fun = new Function(...functionArgNames, expression); // SECURITY!
-        const result = fun(...functionArgValues);
-        return result;
+        const request = {code: expression, argumentNames: functionArgNames, arguments: functionArgValues};
+        if(options?.childProcessJavaScriptEvaluation) {
```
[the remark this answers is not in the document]
dselman: This won't run in the browser. For the browser we could implement something using web workers, but not sure if it is worth it from a security perspective.

src/worker.js
dselman: I believe so, yes - from a Linux perspective. Node privileges do not yet propagate to child processes AFAIK.

src/JavaScriptEvaluator.ts
```diff
@@ -0,0 +1,198 @@
...
+    expireTime: number; // timestamp for when work must be completed
+    request: EvalRequest; // payload to process
+    resolve: (result: any) => void; // promise callback for success
+    reject: (result: any) => void; // promise callback for failure
+}
+
+export type EvalResponse = {
+    result: any // success if not null
+    timeout?: boolean // if true the promise is rejected due to timeout
+    starvaton?: boolean // if true the promise is rejected due to lack of child process
```
dselman, line 41: Typo.

src/JavaScriptEvaluator.ts
```diff
@@ -0,0 +1,198 @@
...
+    private doWork(work: WorkItem, options: EvalOptions): Promise<EvalResponse> {
+        return new Promise((resolve, reject) => {
+            const start = new Date().getTime();
+            // check for browser
+            if (!child_process.fork) {
+                reject({ message: 'Cannot use evalChildProcess because child_process.fork is not defined.' });
+            }
+            // on timeout will send SIGTERM
+            const worker = child_process.fork('./dist/worker.js', { timeout: options.timeout, env: {} });
+            this.workers.push(worker);
```
dselman, line 161: Move below throw.

src/worker.js
```diff
@@ -0,0 +1,44 @@
...
+process.on('message', (msg) => {
+    if (!process.send) {
+        throw new Error('Cannot communicate with parent process!');
+    }
+    // we wrap the code in "use strict" to so the code cannot access the global scope
+    const code = `"use strict";
+    ${msg.code}
+    `;
+    try {
+        // we expose just two node modules to the function: dayjs 'now' and the 'jsonpath' module
```
dselman, line 32: Remove ref to 'now'

dselman (commented):
Minor changes.

PR #52, opened 2026-01-01
fix: add formatters for Duration and Period types

Shubh-Raj:
## Summary
Fixes: https://github.com/accordproject/template-playground/issues/14
When Duration or Period types are used in TemplateMark templates, they were rendering as raw JSON instead of human-readable text.
### Before
The penalty is {"$class":"
org.accordproject.time@0.3.0.Duration
","amount":2,"unit":"days"}.

### After
The penalty is 2 days.

## Changes
- Added `src/drafting/Duration/index.ts` - formats Duration/Period as "{amount} {unit}"
- Registered drafters in `src/drafting/index.ts` (3 lines)
- Added unit tests in `test/DurationDrafter.test.ts`
## Testing
All 53 tests pass.

### DIFF
--- a/src/drafting/Duration/index.ts
+++ b/src/drafting/Duration/index.ts
@@ -18,18 +18,28 @@ type Duration = {
 };
 
 /**
+ * Type guard to check if a value is a valid Duration
+ */
+function isDuration(value: unknown): value is Duration {
+    return (
+        value != null &&
+        typeof value === 'object' &&
+        'amount' in value &&
+        'unit' in value &&
+        typeof (value as Duration).amount === 'number' &&
+        typeof (value as Duration).unit === 'string'
+    );
+}
+
+/**
  * Creates a drafter for Duration/Period types
- * @param {object} value the duration or period
+ * @param {unknown} value the duration or period (validated at runtime)
  * @returns {string} the text (e.g., "2 days")
  */
-export default function durationDrafter(value: Duration): string {
-    // Validate input to avoid runtime errors
-    if (value == null || typeof value !== 'object') {
+export default function durationDrafter(value: unknown): string {
+    if (!isDuration(value)) {
         return '0 unknown';
     }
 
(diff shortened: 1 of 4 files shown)


src/drafting/Duration/index.ts
```diff
@@ -0,0 +1,27 @@
...
+    unit: string;
+};
+
+/**
+ * Creates a drafter for Duration/Period types
+ * @param {object} value the duration or period
+ * @returns {string} the text (e.g., "2 days")
+ */
+export default function durationDrafter(value: Duration): string {
+    return `${value.amount} ${value.unit}`;
```
mttrbrts: Noting that this doesn't scale to international durations yet (although the same problem exists for other compound types). We would need to make the template engine aware of Concerto vocabularies to solve this problem.

Similarly, the ordering is not consistent across locales. 

| Language	| Format Style |	Example (2 Days)	|Note|
|--|--|--|--|
English|	Number + Unit|	2 days|	Pluralizes unit
French|	Number + Unit|	2 jours	|Space is mandatory
Hungarian|	Number + Unit	|2 nap	|Unit is always singular
Arabic	|Dual Form	|يومان	|The number 2 is often built into the word
Chinese	|Number + Counter	|2天	|No spaces between parts

The following code snippet would help if we have the locale available in the template.

```js
// Using Intl.RelativeTimeFormat (Standard in modern JS)
const rtf = new Intl.RelativeTimeFormat('en', { numeric: 'always' });
console.log(rtf.format(2, 'day')); // "in 2 days"
```

mttrbrts (commented):
Please review Copilot suggestions

Shubh-Raj:
Fixed the clone PR issue as discussed in meet.

mttrbrts:
Sorry for the delay in getting back to this one @Shubh-Raj!

Shubh-Raj:
> Sorry for the delay in getting back to this one @Shubh-Raj!

No issues! Thanks for the review.

PR #53, opened 2026-01-01
fix: add type checking for optional properties

Shubh-Raj:
## Description

Add compile-time type checking for optional properties in TemplateMark templates. When an optional property is used without a guard (`{{#optional}}`, `{{#if}}`, or `{{#with}}`), the type-checker now throws an error at compile time instead of causing a runtime failure.

**Before:** Runtime error - `"No values found for path '$['last']'"`
**After:** Compile-time error - `"Optional properties used without guards: last"`

## Changes

- Added validation in [checkTypes()](cci:1://file:///home/shubhraj/OpenSource/template-engine/src/TemplateMarkInterpreter.ts:557:4-630:5) to detect unguarded optional property access
- Uses path-based scoping so guards only apply to descendant nodes
- Throws proper [Error](cci:2://file:///home/shubhraj/OpenSource/template-engine/src/TemplateMarkToJavaScriptCompiler.ts:30:0-34:2) object with `.errors` property for structured access
- Added test case `test/templates/bad/optional-noguard/`

## Testing

All 26 TemplateMarkInterpreter tests pass. All 33 snapshots pass.

## Related Issues

Fixes https://github.com/accordproject/template-playground/issues/11

### DIFF
--- a/src/TemplateMarkInterpreter.ts
+++ b/src/TemplateMarkInterpreter.ts
@@ -575,12 +575,60 @@ export class TemplateMarkInterpreter {
         const serializer = new Serializer(factory, modelManager);
         try {
             serializer.fromJSON(templateMark);
-            return templateMark;
         }
         catch (err) {
             throw new Error(`Generated invalid agreement: ${err}: ${JSON.stringify(templateMark, null, 2)}`);
         }
+
+        const errors: Array<{ propertyName: string, message: string }> = [];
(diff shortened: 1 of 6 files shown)


mttrbrts (changes requested):

Shubh-Raj:
Following up.

mttrbrts (approved):

PR #73, opened 2026-01-26
 Template compilation caching to improve performance

### LINKED ISSUE #42: precompiled trigger support
A suggested improvement to the performance of TemplateArchiveProcessor is to add a separate compile step. While each trigger requires compilation of involving agreement-specific data, there is additional overhead in doing this every time the agreement is triggered. With this suggested logic added the trigger call can optionally skip the precompile step in cases where the code has not changed and has already been compiled. The output of the compilation could be stored for later use and passed as a param when triggering. This could give some flexibility around when the compile happens always at runtime, or if it can be done once on first trigger and then cached, or if this could be done on agreement creation, or possibly at some later APAP event such as an agreement has been signed.

Nabihamasroor:
Fixes #42 

## Changes
- Added `TemplateValidator` class for validating templates before processing
- Implemented caching system using Map to store compiled code and avoid recompilation
- Modified init() and trigger() methods to accept optional **compiledTemplate** parameter
- Added cache key generation based on template content hash
-Implemented cache management with **invalidateCache()** and **clearAllCaches()** methods
- Updated tests to validate caching behavior and performance improvements

## Features
- **Separate Compile Step**: Compile TypeScript once instead of on every trigger/init call
- **Performance Monitoring**: Cache hit/miss statistics for optimization tracking

### DIFF
--- a/src/TemplateArchiveProcessor.ts
+++ b/src/TemplateArchiveProcessor.ts
@@ -1,184 +1,219 @@
-/*
- * Licensed under the Apache License, Version 2.0 (the "License");
- * you may not use this file except in compliance with the License.
- * You may obtain a copy of the License at
- *
- * http://www.apache.org/licenses/LICENSE-2.0
- *
- * Unless required by applicable law or agreed to in writing, software
- * distributed under the License is distributed on an "AS IS" BASIS,
- * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
- * See the License for the specific language governing permissions and
- * limitations under the License.
- */
-
-/* eslint-disable @typescript-eslint/no-explicit-any */
-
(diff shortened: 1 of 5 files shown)


dselman (changes requested):
I like this but:
1. Is the validator a separate concern to the compilation output caching? It looks like it should be its own PR?
2. I want to reimplement the compiler using the code I have in [a draft PR](https://github.com/accordproject/template-engine/pull/50) removing the two slash dependency. Let me complete that work first then you can add caching on top.

mttrbrts:
Thanks for this contribution! Before we can review, please resolve the merge conflicts with the base branch.

```bash
git fetch origin main
git rebase origin/main
# Resolve conflicts
git push --force-with-lease
```

Once the conflicts are resolved, we'll proceed with the review.

---

PR #88, opened 2026-02-10
fix: clarify error message for JavaScript evaluation failure

Sana9058:
### Summary
Improves the error message thrown in `evaluateJavaScript` when required
inputs are missing.

### What changed
- Replaced a generic error message that printed raw `fn` and `data`
- Added a clearer, user-friendly message explaining that template data
  or function source is missing

### Motivation
The previous error message could expose large objects in logs and was
hard to understand during debugging. This change makes the failure
reason explicit without changing any behavior.

### Notes
- No functional or logic changes
- Error condition remains the same


Sana9058:
Thanks for taking a look!

This change only improves the clarity of the error message when JavaScript
evaluation is attempted without required inputs. No behavior or logic was
changed.

Happy to adjust or split commits if you’d prefer a different structure.

Issue #104, opened 2026-03-10
Feature : Template Validation & Static Analysis


Ananya-vastare:
Currently template errors (like missing variables) are only detected during rendering, which makes debugging difficult. I propose adding a template validation feature that checks TemplateMark templates before execution. The validator would analyze the template, extract variable references, and compare them with the Concerto data model. It would report issues such as missing variables, type mismatches, or unused fields. This would improve developer experience and help detect errors earlier in the development process. I’d be happy to work on implementing this along with tests and documentation.

Ananya-vastare:
@dselman @ekarademir @jamieshorten @martinhalford  I am interested on working on this feature . If it’s okay with the maintainers, could this issue be assigned to me?

PR #105, opened 2026-03-10
feat: add template validation and static analysis #104

Ananya-vastare:
#104 
This feature introduces a validation step for TemplateMark templates before rendering.  
It performs static analysis to detect missing variables referenced in the template and fields defined in the data model but unused in the template.  

The goal is to improve developer experience by catching template errors early in the development process.  
This validation helps ensure better consistency between TemplateMark templates and their associated Concerto data models.


Ananya-vastare:
@ekarademir @jamieshorten @dselman @martinhalford please look into this pr thank you

mttrbrts:
Thanks for the contribution! However, this approach doesn't align with the project's design philosophy.

The Accord Project libraries are designed so that validation is handled by Concerto (specifically `@accordproject/concerto-core`), and applications can choose whether to use validation or not. This separation of concerns is intentional.

A few issues with the current approach:
1. The regex-based variable detection (`{{variable}}`) doesn't account for TemplateMark's full syntax (blocks like `{{#if}}`, `{{#optional}}`, `{{#with}}`, etc.)
2. Checking variables against a plain `Record<string, any>` model misses the type information that Concerto provides
3. This duplicates what Concerto's validation layer already handles

If you're interested in validation improvements, I'd suggest looking at how Concerto validation is used elsewhere in the codebase — the goal would be to leverage existing Concerto capabilities rather than adding a parallel validation system.

Ananya-vastare:
Thanks for the feedback — that makes sense, and I see the issue with duplicating Concerto’s responsibilities.

I’ve updated the approach to align with the design philosophy:

1. Removed regex-based template parsing entirely (no {{variable}} scanning)
2. Removed manual Record<string, any> validation logic
3. Delegated all data validation fully to @accordproject/concerto-core using ModelManager + Serializer
4. Treated TemplateMark purely as a rendering concern, not a validation source
5. Any cross-checking between template and data is now optional and treated as warnings only (not validation errors), and does not replicate Concerto’s type system

So now:

1. Concerto is the single source of truth for validation
2. TemplateMark is only used for template structure/rendering
3. The previous parallel validation layer has been removed


Let me know if you'd like me to align further with how other parts of the codebase are handling Concerto validation — happy to adjust.
@mttrbrts

mttrbrts:
@Ananya-vastare Did you forget to push a new commit with the changes described in your message?

mttrbrts:
Closing, but feel free to reopen @Ananya-vastare if you can push the changes that you describe in your comment above

PR #110, opened 2026-03-19
fix: correct PID check, preserve stack traces, and add missing return after queue overflow reject

Swastik-Prakash1:
Fix some edge cases & error handling

Fixed PID check so 0 isn’t treated as missing

Switched to console.error and added stack traces for better debugging

Stopped execution after queue overflow reject (was still pushing to queue)

### DIFF
--- a/src/JavaScriptEvaluator.ts
+++ b/src/JavaScriptEvaluator.ts
@@ -47,6 +47,7 @@ export type EvalResponse = {
     timeout?: boolean // if true the promise is rejected due to timeout
     starvation?: boolean // if true the promise is rejected due to lack of child process
     message?: string; // if promise rejected due to a caught exception this will be set
+    stack?: string;
     elapsed?: number; // the elapsed time in ms to process the work item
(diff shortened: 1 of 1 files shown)


Sana9058 (commented):
Hi! I reviewed this PR, and the changes for PID validation, stack trace handling, and stopping execution after queue overflow look correct and align with the expected behavior.

One small observation: in a few places, errors are logged using console.log(err). It might be better to use console.error(err) for clearer error reporting, as mentioned in earlier comments.

Overall, the implementation looks good to me. Thanks for working on this!

src/JavaScriptEvaluator.ts
```diff
@@ -137,15 +138,17 @@ export class JavaScriptEvaluator {
                         .catch(err => {
                             console.log(err);
```
Sana9058, line 139: One small observation:

In this catch block, console.log(err) is being used. 
Since errors were switched to console.error in other places, 
it might be better to use console.error(err) here as well for consistency.

Let me know what you think.

PR #113, opened 2026-03-21
fix(template): guard optional property in nested test

### LINKED ISSUE #112: Fix: "optional-nested" test fails after stricter optional checks in PR #53
**Description**
After the changes in PR #53, which introduced stricter compile-time checks for optional properties, the `should generate optional-nested` test in `test/TemplateMarkInterpreter.test.ts` is failing with the error: `Optional properties used without guards: age`.

This happens because the template at `test/templates/good/optional-nested/template.md` uses the `age` property, which is correctly marked as `optional` in its model, without an `{{#optional}}` guard.

**To Reproduce**
1. Pull the latest `main` branch.
2. Run `npm install`.
3. Run `npx jest test/TemplateMarkInterpreter.test.ts`.
4. See the failing `should generate optional-nested` test.

**Expected Behavior**
All tests should pass. The `optional-nested` template should be updated to correctly use a guard for the optional `age` property, aligning with the new validation rules.

**Proposed Fix**
In `test/templates/good/optional-nested/template.md`, wrap the `{{age}}` variable in an `{{#optional}}` block to correctly handle the optional property.

Change:
`- You are *{{age}}* years old`

To:
`- You are *{{#optional age}}{{this}}{{/optional}}* years old`

OmarElzero:
This pull request resolves a test failure in `test/TemplateMarkInterpreter.test.ts` that occurred after the stricter optional property checks were introduced in PR #53.

The `should generate optional-nested` test was failing with the error `Optional properties used without guards: age`. This was because the template was using an optional property (`age`) without a guard.

This fix wraps the `age` variable in the `test/templates/good/optional-nested/template.md` file with an `{{#optional}}` block, ensuring it conforms to the new validation rules. The corresponding test snapshot has also been updated.

Fixes #112

### DIFF
--- a/test/__snapshots__/TemplateMarkInterpreter.test.ts.snap
+++ b/test/__snapshots__/TemplateMarkInterpreter.test.ts.snap
@@ -3506,37 +3506,42 @@ exports[`templatemark interpreter should generate optional-nested 1`] = `
           ],
         },
         {
-          "$class": "org.accordproject.commonmark@0.5.0.List",
+          "$class": "org.accordproject.commonmark@0.5.0.Paragraph",
           "nodes": [
             {
-              "$class": "org.accordproject.commonmark@0.5.0.Item",
+              "$class": "org.accordproject.ciceromark@0.6.0.Optional",
+              "elementType": "Integer",
+              "hasSome": true,
+              "name": "age",
               "nodes": [
                 {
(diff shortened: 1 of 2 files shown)


mttrbrts:
Unfortunately this was already fixed by another contributor. Thanks anyway!

PR #118, opened 2026-03-22
fix: correctly evaluate clause and conditional expressions

### LINKED ISSUE #117: Bug: {{#clause condition}} logic is bypassed and incorrectly evaluates false as true
## Overview
The {{#clause name condition="..."}} feature fails to evaluate conditions correctly due to two related bugs in TemplateMarkInterpreter.ts. This affects both conditional clauses and {{#if}} constructs (e.g., in the template playground).

## The Bugs
1. Incorrect Evaluation Order (Bypassed Logic)
Implicit undefined checks run before the explicit condition. If a clause checks an optional field that happens to be undefined, the clause is skipped immediately, completely bypassing the else if (context.condition) branch and never evaluating the user's condition logic.

2. Boolean Parsing Error ("false" === true)
Condition results are stored as JSON-stringified values ("true" / "false"). The interpreter checks truthiness using !!result. Because a non-empty string like "false" is truthy, !!"false" incorrectly evaluates to true.

Impact: Any condition returning false is treated as true, causing clauses to render when they should be hidden.

## Steps to Reproduce

for the first bug
the clause with condition won’t work if the optional field is missing.

Image

for the second bug
Create a template explicitly returning false: `{{#clause address condition="return false"}} ... {{/clause}}`

Provide JSON data where the address field exists and is populated.

Generate the agreement.

Image

Expected: The address clause is NOT rendered.
Actual: The address clause IS rendered because !!"false" evaluates to true.

salama968:
# Closes #117 

This PR resolves issue #117 by fixing two critical bugs in TemplateMarkInterpreter.ts that prevented {{#clause condition}} statement from evaluating correctly.

## changes
- Fixed Evaluation Precedence: Reordered the logic in clause definition handling so that explicit user conditions (context.condition) are evaluated before the implicit undefined field checks. This ensures the user's condition logic is no longer bypassed when optional fields are missing.

- Fixed Boolean Parsing: Replaced the flawed !!result truthiness check with JSON.parse(resultStr). Previously, a condition returning false was stringified as "false", which incorrectly evaluated to true. It now correctly parses to a boolean false.

## Tests
Added 3 new Jest test cases (clause-condition-*) to explicitly verify:

- Clauses with conditions properly hide when optional fields are missing.

- Clauses with conditions properly render when optional fields are present.

- Clauses with explicit return false conditions do not render, even if the data field exists (verifying the boolean parsing fix).

## related issues
- issue #117


mttrbrts (approved):
❤️ nice fix, thanks! 

@copilot revert the change to the src/runtime/declarations.ts file

PR #123, opened 2026-03-25
 [feat] LLM Based Template Logic Executor

devanshi00:
### Overview
This PR introduces an **LLM-based generic logic executor** for Accord Project templates. It enables execution of templates even when no `logic.ts` is provided, using a reasoning-based fallback powered by an LLM.

The goal is to explore a flexible, extensible alternative to traditional TypeScript-based logic while maintaining full backward compatibility.

- **LLM Executor**
  - Acts as a fallback or replacement for `logic.ts`
  - Processes:
    - contract text  
    - Concerto models  
    - current state  
    - incoming request  
  - Produces:
    - updated state  
    - response object  
    - emitted events  

- **Execution Modes**
  - `disabled` – default behavior (TypeScript only)  
  - `fallback` – use LLM only if `logic.ts` is absent/fails  
  - `force` – always use LLM executor  

- **Integration**
  - Integrated into `TemplateArchiveProcessor`
  - Seamlessly fits into existing execution pipeline  

- **Normalization Layer**
  - Ensures outputs conform to Accord runtime expectations:
    - timestamps  
    - identifiers  
    - ordering  

- **Backward Compatibility**
  - Existing templates with `logic.ts` remain unaffected  

- **Logging**
  - Clear indication of execution path:
    - TypeScript vs LLM 

## PR Review 
@DianaLease @dselman


mttrbrts:
Thanks for this contribution! Before we can review, please resolve the merge conflicts with the base branch and fix the DCO sign-off.

```bash
git fetch origin main
git rebase origin/main --signoff
# Resolve conflicts
git push --force-with-lease
```

Once the conflicts are resolved and commits are signed off, we'll proceed with the review.

---

devanshi00:
@mttrbrts @DianaLease @dselman  added the llm executor with json schema for anthropic and openai provider. Also added a stateful and stateless templates results to check the feasibility and correctness of the llm executor.

mttrbrts (commented):
> **Note:** This review was generated with the assistance of Claude (AI). The comments reflect the maintainer's views but were drafted and posted by an AI code review tool.

---

Hi @devanshi00 👋 — really appreciate the effort here! An LLM-based fallback executor is a genuinely exciting direction for Accord Project, and the core design (three modes, schema-enforced structured outputs, provider abstraction) is solid. Thanks for tackling this!

Before we can merge, there are a few things we'd like to work through together — some are blockers, and some are things we'd love to discuss to find the best path to land this incrementally. I've left inline comments on the specifics, but the high-level themes are:

**Blockers for any merge:**
- The PR's head branch appears to be `main`, which means it likely has unresolved merge conflicts with upstream. Could you rebase onto the current `main` and resolve those before we do a deeper review?
- The project README has been fully replaced with LLM Executor-specific content. The existing README is the public face of this project — the LLM executor is a great *addition*, not a replacement. Let's keep the original content and add a new section (or a `docs/llm-executor.md`) for this feature.

**Things to discuss for incremental merging:**
- The PR is currently 233 files / ~15k lines, which makes it very hard to review safely and increases the risk of merge conflicts going forward. We'd love to find a way to land the core feature (`src/llm/` + the `TemplateArchiveProcessor` integration) first, and follow up with examples in a separate PR.
- The four LLM SDK packages (`openai`, `@anthropic-ai/sdk`, `@google/genai`, `@mistralai/mistralai`) are currently in `dependencies`, which means every downstream consumer of `template-engine` installs them even if they never use the LLM executor. We should chat about moving these to `optionalDependencies` or `peerDependencies` so the package stays lightweight for users who don't need this feature.
- The package version jumped from `2.7.1` → `3.0.1` and also upgrades several `@accordproject/*` packages to new major versions. Were those dependency upgrades intentional as part of this PR? If so, it might be cleaner to land them separately so the diff is easier to follow.

None of this should block the conversation — please don't be discouraged! The core idea is great and we want to get it merged. Happy to jump on a call if it would help work through the path forward. 🙏

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@accordproject/template-engine",
-  "version": "2.7.1",
+  "version": "3.0.1",
   "description": "Generation of AgreementMark from TemplateMark + JSON Data",
```
mttrbrts, line 4: Bumping to `3.0.1` is quite a jump from `2.7.1`! A major semver bump tells downstream users there's a breaking change. Since the `TemplateArchiveProcessor` constructor change is backward-compatible (the new `llmConfig` param is optional), we'd normally keep this as a minor bump (`2.8.0`). Could you clarify whether the major bump is intentional? If it's tied to the `@accordproject/*` dep upgrades, it might be cleaner to land those in a separate PR so each change is easy to reason about independently.

package.json
```diff
@@ -39,23 +39,28 @@
...
-    "@accordproject/markdown-template": "0.17.2",
+    "@accordproject/cicero-core": "0.26.0",
+    "@accordproject/concerto-codegen": "4.0.1",
+    "@accordproject/concerto-core": "4.0.3",
+    "@accordproject/concerto-util": "4.0.3",
+    "@accordproject/concerto-vocabulary": "4.0.3",
+    "@accordproject/markdown-common": "0.18.0",
+    "@accordproject/markdown-template": "0.18.0",
+    "@anthropic-ai/sdk": "^0.100.1",
+    "@google/genai": "^2.8.0",
```
mttrbrts: These four LLM SDK packages are large and have their own transitive dependency trees. Adding them to `dependencies` means every consumer of `template-engine` — including those who only use the existing TemplateMark drafting pipeline and never touch the LLM executor — will install all of them unconditionally.

Two options worth discussing:
1. **`optionalDependencies`** — npm installs them by default but won't fail if they're absent, and bundlers can tree-shake them. Requires dynamic `import()` in `Reasoners.ts` with a helpful error if the package isn't found.
2. **`peerDependencies`** — the caller installs only what they need. Requires a bit more documentation but keeps the package lean.

What do you think? Option 1 is probably the least disruptive for the PR as it stands, and we could tighten it further in a follow-up.

src/TemplateArchiveProcessor.ts
```diff
@@ -185,4 +211,80 @@ export class TemplateArchiveProcessor {
...
+    async trigger(
+        data: any,
+        request: any,
+        state?: any,
+        currentTime?: string,
+        utcOffset?: number
+    ): Promise<TriggerResponse> {
+
+        console.log(`\n[TemplateArchiveProcessor.trigger]`);
+        console.log(`Mode: ${this.llmConfig?.mode ?? 'disabled'}`);
```
mttrbrts: These `console.log` statements will fire on every `trigger()` and `init()` call, even for users who haven't configured an LLM at all (the `disabled` default). That's going to be noisy for existing consumers who upgrade.

Since `LLMExecutorConfig` already has a `verbose` flag, these could be gated on `this.llmConfig?.verbose` — or moved entirely into `LLMExecutor` where the `verbose` flag is already used consistently. The `TemplateArchiveProcessor` itself probably shouldn't log anything unconditionally.

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,598 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Template } from '@accordproject/cicero-core';
+import { BaseReasoner, ChatMessage, JsonSchema, createReasoner } from './Reasoners';
+import { LLMExecutorConfig } from './LLMConfig';
+ 
+let _schemaCache: { definitions: Record<string, any> } | null = null;
+let _schemaPath = ''; // empty = not set yet
```
mttrbrts: Module-level mutable globals are a bit fragile here — if someone constructs two `LLMExecutor` instances for different templates (each with their own `schema.json`), the second `setSchemaPath()` call silently invalidates the cache for the first. It also makes the API slightly surprising: the caller has to remember to call `setSchemaPath()` *before* constructing the executor, with no enforcement at the type level.

One alternative worth considering: pass the schema path (or the resolved schema object) directly in `LLMExecutorConfig`:

```typescript
export interface LLMExecutorConfig {
  mode: LLMMode;
  provider: LLMProviderConfig;
  schemaPath?: string;   // or: schema?: { definitions: Record<string, any> }
  verbose?: boolean;
}
```

That way each executor is fully self-contained and the contract is clear at construction time. Happy to discuss — there may be a reason for the current approach (e.g. Anthropic grammar cache requiring the same schema object reference) that I'm missing!

src/llm/Reasoners.ts
```diff
@@ -0,0 +1,430 @@
...
+    > & { apiKey?: string },
+    baseURL: string,
+    defaultApiKey = 'placeholder'
+  ) {
+    super();
+    this.client = new OpenAI({
+      apiKey: config.apiKey || defaultApiKey,
+      baseURL,
+      dangerouslyAllowBrowser: true,
+    });
```
mttrbrts: `dangerouslyAllowBrowser: true` suppresses a security warning from the OpenAI SDK that exists for good reason — sending API keys from a browser exposes them to users via DevTools. If browser support isn't a goal for the LLM executor right now, this flag shouldn't be needed at all (the executor would only run server-side). If browser support *is* intentional, it needs a proxy layer so the key never reaches the client. Worth flagging so we make a deliberate decision here rather than inheriting it silently!

src/index.ts
```diff
@@ -17,4 +17,5 @@ import type { TemplateLogic } from './slc/SmartLegalContract.d.ts';
 export { TemplateMarkInterpreter } from './TemplateMarkInterpreter';
 export { TemplateArchiveProcessor } from './TemplateArchiveProcessor';
 export { TemplateLogic }
+export { setSchemaPath } from './llm/LLMExecutor';
 export * from './utils';
```
mttrbrts, line 20: Exporting `setSchemaPath` from the package root makes it part of the public API, which means we'd have to support it as a breaking change if we ever want to remove it. If we move `schemaPath` into `LLMExecutorConfig` (as suggested in the comment on `LLMExecutor.ts`), this export goes away entirely and the public API surface stays cleaner. Worth discussing as part of the same design question!

dselman:
@devanshi00 this is impossible to review. I suspect most of these modifications are formatting changes that your editor config has applied. Get this down to < 20 files that have truly been modified and that are tightly focused on the problem that you are trying to solve. Little and often is much easier to review than Big Bang.

devanshi00:
> @devanshi00 this is impossible to review. I suspect most of these modifications are formatting changes that your editor config has applied. Get this down to < 20 files that have truly been modified and that are tightly focused on the problem that you are trying to solve. Little and often is much easier to review than Big Bang.

Actually  I added complete 8 examples from cierco-template-library which is making it so huge. I will remove them and just suggest a way to get outputs as I got.

src/llm/Reasoners.ts
devanshi00: Yes I removed it.

src/TemplateArchiveProcessor.ts
devanshi00: Removed them

package.json
devanshi00: I think the current versions of accordproject as per the latest commit and mine match.
Please let me know if there is any issue.
Marking it resolved for now

package.json
devanshi00: Done, used option 1.

DianaLease (approved):
Great work! As mentioned on our call today, you can iterate on this in the next PR to include tests and output validation against concerto model. I'll wait for @dselman to review as well before merging.

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+import { Template } from '@accordproject/cicero-core';
+import { BaseReasoner, ChatMessage, JsonSchema, createReasoner } from './Reasoners';
+import { LLMExecutorConfig } from './LLMConfig';
+import {
+  getRootTypes,
+  isStatelessTemplate,
+  treeShakeModel,
+} from './ModelManagerSchema';
+
+export type TriggerResponse = {
```
dselman: You should import the types from here https://github.com/accordproject/template-engine/blob/7667cbe97dd6f458106655457a14743ea3a0fa05/src/TemplateArchiveProcessor.ts#L31

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+  state: object;
+  events: object[];
+};
+ 
+export type InitResponse = {
+  state: object;
+};
+ 
+// Schema helpers 
+function enforceAdditionalPropertiesFalse(schema: Record<string, any>): Record<string, any> {
```
dselman, line 76: This looks like it might be a bug or a feature that needs to get pushed down into the JSONSchemaVisitor. Can you please create an issue for the codegen repo so we can enable this via a flag?

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+    }
+    return result;
+  }
+  return node;
+}
+ 
+// Strips JSON-Schema keywords that strict structured-output APIs reject and
+// pins the Concerto `$class` discriminator to a `const` so the model emits the
+// correct fully-qualified type name.
+const UNSUPPORTED_KEYWORDS = new Set([
```
dselman, line 128: This is interesting... Again, perhaps this should live in the JSONSchemaVisitor (flags to skip the code that adds there)?

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+  'minLength',
+  'maxLength',
+  'minimum',
+  'maximum',
+  'exclusiveMinimum',
+  'exclusiveMaximum',
+  'multipleOf',
+]);
+
+function cleanForStructuredOutput(node: any): any {
```
dselman, line 147: Ideally we would not need this - the output from the JSONSchemaVisitor would be "clean" enough for usage by LLMs.

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+  return node;
+}
+
+/**
+ * Fully resolves a single tree-shaken type into a self-contained, strict
+ * JSON Schema: inlines every `$ref`, forces `additionalProperties: false`, and
+ * strips provider-unsupported keywords.
+ */
+function resolveTypeSchema(
+  definitions: Record<string, any>,
```
dselman: It is not clear what definitions arg and the return value should contain. Add some docs please or better typing.

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+function resolveUnionSchema(
+  definitions: Record<string, any>,
+  fqns: string[]
+): Record<string, unknown> | null {
+  if (fqns.length === 0) return null;
+  if (fqns.length === 1) return resolveTypeSchema(definitions, fqns[0]);
+  return { anyOf: fqns.map(fqn => resolveTypeSchema(definitions, fqn)) };
+}
+
+/** Attach a description; only stamp `additionalProperties:false` on plain object
```
dselman: This function seems to be doing two distinct things. Should they be separated?

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+    type: 'object',
+    properties,
+    required,
+    additionalProperties: false,
+  };
+}
+ 
+/** Returns true for providers that enforce schemas natively and need the full
+ *  definition expanded from schema.json (OpenAI, Anthropic). */
+function usesFullSchema(config: LLMExecutorConfig): boolean {
```
dselman: Sounds like this should live on the provider interface.

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+    return JSON.parse(raw);
+  } catch {
+    const match = raw.match(/```(?:json)?\s*([\s\S]*?)```/i);
+    if (match) return JSON.parse(match[1]);
+    throw new Error(`LLM did not return valid JSON. Raw output: ${text}`);
+  }
+}
+
+function assertInitShape(value: any): asserts value is InitResponse {
+  if (!value || typeof value !== 'object' || !value.state || typeof value.state !== 'object') {
```
dselman, line 406: TODO - replace these with Concerto based deserialization:
template.getModelManager().getSerializer().fromJSON( json ) <-- throws if object is not a valid Concerto instance

src/llm/LLMExecutor.ts
```diff
@@ -0,0 +1,621 @@
...
+  const rawState = response.state;
+  const identifier: string =
+    rawState?.$identifier ||
+    data?.$identifier ||
+    data?.clauseId ||
+    data?.contractId ||
+    'state-1';
+  
+  function insertAfterClass(obj: Record<string, any>, key: string, value: string): Record<string, any> {
+    if (!obj || typeof obj !== 'object' || Array.isArray(obj)) return obj;
```
dselman: Not sure why you care about the ordering of the properties - this is not enforced by Concerto.

src/llm/ModelManagerSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Concrete state types declared in the template's model.
+ *
+ * Two signals are combined, because templates declare state inconsistently:
+ *   1. subclasses of the runtime `State` base (`template.getStateTypes()`); and
+ *   2. concepts whose name ends with `State` in the Accord Project convention
+ */
+export function getStateTypes(template: any): string[] {
+  const fromBase = templateOnly(template.getStateTypes?.());
```
dselman: Use https://github.com/accordproject/template-archive/blob/feb5ec9e27e9364eaf1be595b31e21f28e1c6054/packages/cicero-core/src/template.ts#L476

src/llm/ModelManagerSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+
+/**
+ * Concrete event types declared in the template's model.
+ *
+ * `template.getEmitTypes()` only finds subclasses of the runtime `Obligation`
+ * base, so a template that emits a plain `event` (e.g. `PaymentObligationEvent`,
+ * which doesn't extend `Obligation`) reports nothing. We instead enumerate every
+ * concrete event declaration.
+ */
+export function getEventTypes(template: any): string[] {
```
dselman: Use https://github.com/accordproject/template-archive/blob/feb5ec9e27e9364eaf1be595b31e21f28e1c6054/packages/cicero-core/src/template.ts#L468

src/llm/ModelManagerSchema.ts
```diff
@@ -0,0 +1,177 @@
...
+  return Array.from(new Set([...fromBase, ...byName]));
+}
+
+/**
+ * Resolve the request / response / state / event type names a template declares.
+ * Request and response come from the (reliable) runtime-base accessors; state
+ * additionally falls back to the `*State` naming convention; events are
+ * enumerated from event declarations.
+ */
+export function getRootTypes(template: any): TemplateRootTypes {
```
dselman: Use https://github.com/accordproject/template-archive/blob/feb5ec9e27e9364eaf1be595b31e21f28e1c6054/packages/cicero-core/src/template.ts#L452

https://github.com/accordproject/template-archive/blob/feb5ec9e27e9364eaf1be595b31e21f28e1c6054/packages/cicero-core/src/template.ts#L460

src/TemplateArchiveProcessor.ts
```diff
@@ -44,14 +48,19 @@ export type InitResponse = {
  */
 export class TemplateArchiveProcessor {
     template: Template;
-    private compiledLogicCache?: Record<string, TwoSlashReturn>;
+    private compiledLogicCache?: Record<string, TwoSlashReturn>; // kept from file
+
+    // optional LLM config (added from pasted version)
```
dselman: What does this comment mean?

src/TemplateArchiveProcessor.ts
```diff
@@ -72,18 +81,13 @@ export class TemplateArchiveProcessor {
         const engine = new TemplateMarkInterpreter(modelManager, {});
         const templateMarkTransformer = new TemplateMarkTransformer();
         const templateMarkDom = templateMarkTransformer.fromMarkdownTemplate(
-            { content: this.template.getTemplate() }, modelManager, templateKind);
+            { content: this.template.getTemplate() }, modelManager, templateKind); // kept from file (no { options } arg)
```
dselman: Clean up comment

src/TemplateArchiveProcessor.ts
```diff
@@ -117,70 +121,146 @@ export class TemplateArchiveProcessor {
...
-     * @param {[string]} currentTime - the current time, defaults to now
-     * @param {[number]} utcOffset - the UTC offset, defaults to zero
-     * @param {boolean} [enableCompiledLogicCache] - whether to use the compiled logic cache
-     * @returns {Promise<TriggerResponse>} the response and any events
-     */
-    async trigger(data: any, request: any, state?: any, currentTime?: string, utcOffset?: number, enableCompiledLogicCache?: boolean): Promise<TriggerResponse> {
-        const compiledCode = await this.compileLogic(enableCompiledLogicCache);
+    // Helper: check if explicit TS logic exists (added from pasted version)
+    private hasTypeScriptLogic(): boolean {
+        const logicManager = this.template.getLogicManager();
```
dselman: Use https://github.com/accordproject/template-archive/blob/feb5ec9e27e9364eaf1be595b31e21f28e1c6054/packages/cicero-core/src/template.ts#L484

src/TemplateArchiveProcessor.ts
```diff
@@ -117,70 +121,146 @@ export class TemplateArchiveProcessor {
...
+            return false;
+        }
+        if (logicManager.getLanguage() !== 'typescript') {
+            return false;
+        }
+        const tsFiles: Array<Script> = logicManager.getScriptManager().getScriptsForTarget('typescript');
+        return tsFiles.length > 0;
+    }
+
+    // Helper: check if LLM fallback is enabled (added from pasted version)
```
dselman: Please clean up these comments - we want jsdoc style comments

src/TemplateArchiveProcessor.ts
```diff
@@ -117,70 +121,146 @@ export class TemplateArchiveProcessor {
...
+     * @param {object} request - the request to send to the template logic
+     * @param {object} state - the current state of the template
+     * @param {[string]} currentTime - the current time, defaults to now
+     * @param {[number]} utcOffset - the UTC offset, defaults to zero
+     * @param {boolean} [enableCompiledLogicCache] - whether to use the compiled logic cache
+     * @returns {Promise<TriggerResponse>} the response and any events
+     */
+    async trigger(data: any, request: any, state?: any, currentTime?: string, utcOffset?: number, enableCompiledLogicCache?: boolean): Promise<TriggerResponse> {
+        // FORCE mode — skip TS logic entirely
+        if (this.llmConfig?.mode === 'force') {
```
dselman: A useful feature for testing would be the ability to run BOTH the typescript and the LLM and then compare the results, throwing an exception if they are different.

dselman: I also don't like that makeLLMExecutor is also called below. I think you can simplify the code to only have a single call to makeLLMExecutor; either because there is no ts or because force is on.

src/TemplateArchiveProcessor.ts
```diff
@@ -117,70 +121,146 @@ export class TemplateArchiveProcessor {
...
+        if (this.hasTypeScriptLogic()) {
+            if (enableCompiledLogicCache) {
+                await this.compileLogic(true);
+            }
+            return this.executeTypeScriptInit(data, currentTime, utcOffset);
+        }
+
+        // Fallback to LLM
+        if (this.shouldUseLLM()) {
+            return this.makeLLMExecutor().init(data, currentTime, utcOffset);
```
dselman, line 278: Same issue with the duplicate calls to makeLLMExecutor

.env.example
```diff
@@ -0,0 +1,6 @@
+# Copy this file to `.env` and fill in your keys. `.env` is gitignored.
+# Usage (Node 20.6+, no dotenv dependency needed):
+#   node --env-file=.env copyright-license/run.js --mode force --provider anthropic --model ./copyright-license/model/model.cto
+OPENAI_API_KEY=sk-...
```
dselman, line 5: This will only work when running from Node. TODO - we need to reconcile this with how Playground configures LLMs and make sure they can share config.

webpack.config.js
```diff
@@ -74,6 +74,11 @@ module.exports = {
         // (the default in-process evaluator is used instead). Ignore the self-reference so
         // the bundle builds.
         new webpack.IgnorePlugin({ resourceRegExp: /^@accordproject\/template-engine$/ }),
+        // The LLM SDKs are optional, node-oriented dependencies loaded via dynamic
+        // import() in src/llm/Reasoners.ts. Keep them out of the browser bundle — the
+        // dynamic import rejects at runtime and the reasoner's try/catch surfaces a
+        // helpful "install this package" error instead.
+        new webpack.IgnorePlugin({ resourceRegExp: /^(openai|@anthropic-ai\/sdk)(\/.*)?$/ }),
```
dselman, line 81: Ok - this means this code will not run in Playground...

dselman (changes requested):
Good stuff here. The changes I'd like to see before merge are:
1. Using ModelManager to find derived types (State/Request/Response/Event)
2. Using the methods on the Template for things like hasLogic so we don't duplicate that

I will leave it to your judgement how much of the rest you want to handle in a follow-up PR.

src/index.ts
devanshi00: Changed this . Now using Model Tree Shaking for this utility.

webpack.config.js
devanshi00: Yes, these SDKs are not browser-friendly by default because they’re designed for server-side use, they intentionally guard against exposing API keys in the browser, and they add extra bundle/runtime dependencies that are awkward in a UMD web build.

Two possible designs here:
1. Introduce a small Transport interface so LLMExecutor/Reasoners stay responsible for prompt building, schema handling, retries, and response parsing, while the actual provider call is delegated to a pluggable transport. That lets us keep an SDK-based Node transport, and add separate browser or proxy transports later without changing executor logic.

2. For Playground/browser support, move LLM execution behind a backend service. The browser would only send a normalized payload such as model/messages/schema to the service, and the service would call OpenAI/Anthropic server-side. This keeps API keys out of the browser and avoids pulling provider-specific SDK/runtime concerns into the UMD bundle.

src/llm/ModelManagerSchema.ts
devanshi00: **State** needs two signals because templates declare it inconsistently: some extend the runtime `State` base (caught by `getStateTypes()`), others just name a concept `*State` by convention with no inheritance (missed by `getStateTypes()`, caught only by the `byName` fallback). Neither signal alone covers all real cases, so we union both.
So I have used the current workaround in this code.

src/llm/ModelManagerSchema.ts
devanshi00: `getEmitTypes()`  function which the template exposes which finds events only by inheritance, it checks whether a declaration extends the runtime  base class. 
For instance for an example with  `Obligation` base class, our template's event, `PaymentObligationEvent`, is declared with the `event` keyword but does **not** extend `Obligation`, so `getEmitTypes()` returned `[]` and the event was silently dropped from the schema.

That is why this workaround is required for events also.

src/llm/LLMExecutor.ts
devanshi00: Done

src/llm/LLMExecutor.ts
devanshi00:
```typescript
export interface OpenAIProviderConfig extends BaseProviderConfig {
  provider: 'openai';
  readonly supportsStructuredOutput: true; // ← add
}
```

You want something like this?

src/llm/LLMExecutor.ts
devanshi00: https://github.com/accordproject/concerto-codegen/issues/255

Done. You can check it.

src/llm/LLMExecutor.ts
devanshi00: https://github.com/accordproject/concerto-codegen/issues/253
Done. You can check it.

src/llm/LLMExecutor.ts
devanshi00: Yeah this will be done after the issues get resolved.

src/llm/LLMExecutor.ts
devanshi00: This will be taken care in follow up pr.

.env.example
devanshi00: Yes I will look into that.

src/TemplateArchiveProcessor.ts
devanshi00: Removed the duplicate calls as you suggested

src/llm/LLMExecutor.ts
devanshi00: Done. You can check it 
```typescript
/**
 * A JSON Schema `definitions` map as produced by the tree-shaker: keyed by
 * fully-qualified Concerto type name, each value is the JSON Schema fragment for
 * that type. Fragments may contain `$ref`s pointing at sibling definitions
 * (`#/definitions/<fqn>`), which `deepResolve` later inlines.
 */
type SchemaDefinitions = Record<string, Record<string, any>>;
```

src/llm/LLMExecutor.ts
devanshi00: Seperated them for better modular structure.

src/llm/ModelManagerSchema.ts
devanshi00: Same problem. Using the workaround for state and events.

src/TemplateArchiveProcessor.ts
devanshi00: Removed it

src/llm/LLMExecutor.ts
devanshi00: Yeah I removed it now.

webpack.config.js
devanshi00: You can let me know what you think about this.

src/TemplateArchiveProcessor.ts
devanshi00: Changed the dulplicate calls for LLMExecutor.

devanshi00: > A useful feature for testing would be the ability to run BOTH the typescript and the LLM and then compare the results, throwing an exception if they are different.

I will write tests and enable this mode in follow up pr.

src/TemplateArchiveProcessor.ts
devanshi00: Done.

src/llm/LLMExecutor.ts
dselman: Yes - or isStructuredOutputSupported()

src/llm/ModelManagerSchema.ts
dselman: We should force templates to extend the runtime State type. If they do not they are "stateless".

src/llm/ModelManagerSchema.ts
dselman: We should delegate to the logic in the template. If we want to change it -- we should update it there. Do not add "workarounds" in the template-engine, instead create issues on the upstream repo you are depending on.

src/llm/ModelManagerSchema.ts
dselman: No workarounds here please.

webpack.config.js
dselman: We want to keep Playground browser only. Playground already has access to the LLM keys via settings, and Playground is already calling them for its AI Assistance features. Let's keep as is for now and revist as a separate PR (making template-engine / LLM Executor browser compatible and share settings with Playground.

src/llm/LLMExecutor.ts
```diff
@@ -316,10 +407,17 @@ function assertTriggerShape(value: any, stateless = false): asserts value is Tri
...
+ * assigned. `$identifier` resolves via:
+ * `state.$identifier → data.$identifier → data.clauseId → data.contractId → 'state-1'`.
+ * @param response - the LLM response to enrich, mutated in place
+ * @param timestamp - the ISO timestamp to stamp, defaults to now
+ * @param data - the contract data, used to resolve the state `$identifier`
+ * @returns the same response, with metadata applied
+ */
 function injectRuntimeMetadata<T extends { state?: any; result?: any; events?: any[] }>(
   response: T,
   timestamp: string = new Date().toISOString(),
```
dselman: We need to be able to set the value for "now" via an arg or option for testing.

src/TemplateArchiveProcessor.ts
```diff
@@ -126,33 +135,50 @@ export class TemplateArchiveProcessor {
...
         }
-        return new LLMExecutor(this.template, this.llmConfig);
+        if (!this.llmExecutor) {
+            this.llmExecutor = new LLMExecutor(this.template, this.llmConfig);
+        }
+        return this.llmExecutor;
     }
 
-    // Private TS trigger — delegates to compileLogic() (kept file's logic, not the inline duplicate from pasted)
+    // /**
```
dselman: Clean up please

dselman (changes requested):
Getting closer! The timestamp/now issue we can push to next PR if you prefer. Please remove the workaround however.

src/llm/LLMExecutor.ts
devanshi00: Changed it

src/llm/LLMExecutor.ts
devanshi00: Will do it in next pr

src/llm/ModelManagerSchema.ts
devanshi00: https://github.com/accordproject/template-archive/pull/937

Check this pr
It fixes the workarounds I used. I updated the current code also.

src/llm/ModelManagerSchema.ts
devanshi00: Check the pr I mentioned above

src/llm/ModelManagerSchema.ts
devanshi00: Check the pr I mentioned above

src/llm/LLMExecutor.ts
```diff
@@ -30,6 +26,41 @@ import type { TriggerResponse, InitResponse } from '../TemplateArchiveProcessor'
...
+function getTemplateRoots(template: Template): TemplateRootTypes {
+  return {
+    requests: template.getCustomRequestTypes?.() ?? [],
+    responses: template.getCustomResponseTypes?.() ?? [],
+    states: template.getCustomStateTypes?.() ?? [],
+    events: template.getEmitTypes?.() ?? [],
+  };
+}
+
+/** A template is stateless when it reports no custom state type. */
```
dselman: Is this really necessary?

src/llm/LLMExecutor.ts
```diff
@@ -30,6 +26,41 @@ import type { TriggerResponse, InitResponse } from '../TemplateArchiveProcessor'
...
+
+/**
+ * Resolve the request / response / state / event type names a template
+ * declares. These are reliable — derived from the runtime base classes the
+ * types extend, not from their names — which lets the tree-shaker find them
+ * without guessing based on type naming conventions (e.g. `PayOut` vs `Payout`).
+ */
+function getTemplateRoots(template: Template): TemplateRootTypes {
+  return {
+    requests: template.getCustomRequestTypes?.() ?? [],
```
dselman: Will have to be aligned with template archive. No more "custom"

src/llm/LLMExecutor.ts
```diff
@@ -30,6 +26,41 @@ import type { TriggerResponse, InitResponse } from '../TemplateArchiveProcessor'
  */
 type SchemaDefinitions = Record<string, Record<string, any>>;
 
+/**
+ * The request / response / state / event type names a template declares.
+ * All root-type decisions come from the Template API.
+ */
+interface TemplateRootTypes {
```
dselman, line 33: Do we need this -- why not just call the methods on template?

dselman (changes requested):

devanshi00:
The new commit does the following changes
-  Adds multiprovider support to the LLM Executor
- Resolves the timestamp issue and the isStructuredOutputSupported to live on provider interface(by default it is False if not mentioned in the runtime)
- Migrates to cierco version  2.0.0 for the new isStateful method
- Removed the redundant log of model files from model manager schema
- Refactored the LLM executor code

devanshi00:
TO DO for the next PR:
- Check if we can add the cleanForStructuredOutput() function in the concerto\codegen repo. Check its neccessity also.
- Add concerto validation tests
- Check for Stateful templates(if state is preserved or not)
- Check for integration in the playgorund

src/llm/LLMExecutor.ts
devanshi00: Done according to cierco-core-v2

dselman (approved):

PR #128, opened 2026-03-28
fix: throw Error instead of raw array for TypeScript compilation fail…

### LINKED ISSUE #127: fix: TypeScript compilation errors thrown as raw array instead of Error object
## Problem
When a template formula contains invalid TypeScript, `TemplateMarkToJavaScriptCompiler.compile()` throws a raw `Array<CompilerError>` instead of an `Error` object.

This means any caller that catches this gets no `.message`, no stack trace just `[object Object]` when displayed.

The rich error information (renderedMessage, line, character) is collected correctly by twoslash but is completely inaccessible to the caller.

## Reproduction
Use the existing template fixture `test/templates/bad/formula-no-method` the thrown value is a raw array as seen in the snapshot.

## Fix
Wrap the compilation errors into a proper `Error` object with a human-readable message including node ID, error description, line and column number.

## Related
`TemplateArchiveProcessor` has a similar issue where `result.errors` is not checked after compilation. Will address in a follow-up issue.

Sup-ri-tha:
Fixes #127 

## Problem
When a template formula contains invalid TypeScript, TemplateMarkToJavaScriptCompiler throws a raw Array<CompilerError> instead of an Error object. This means callers get no .message, no stack trace just [object Object].

## Fix
Wrap the compilation errors into a proper Error with a human-readable message including the node ID, error description, line and column number.

## Notes
- 2 pre-existing test failures unrelated to this change:  `optional-nested` and `JavaScriptEvaluator` stress test. Verified both fail on main without this fix.
- TemplateArchiveProcessor has a related issue where result.errors is not checked after compilation. Will address this in a follow-up PR once this is merged.

### DIFF
--- a/src/TemplateMarkToJavaScriptCompiler.ts
+++ b/src/TemplateMarkToJavaScriptCompiler.ts
@@ -105,7 +105,11 @@ export class TemplateMarkToJavaScriptCompiler {
             return compiled;
         }
         else {
-            throw errors;
+            const messages = errors.map(e => {
+                const errorList = e.errors.map(d => `${d.renderedMessage} (line ${d.line}, col ${d.character})`);
+                return `In '${e.nodeId}': ${errorList.join(', ')}`;
+            });
+            throw new Error(messages.join('\n'));
         }
     }
 }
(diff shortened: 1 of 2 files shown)


mttrbrts:
Closing as stale. If you want to address the comments from the review, please reopen the PR. Thanks.

PR #130, opened 2026-03-28
fix: use cross-platform Jest path in test and coverage scripts

### LINKED ISSUE #129: bug: npm test fails on Windows - Jest invoked via bash shim
## Problem

`npm test` fails immediately on Windows with a SyntaxError:

```
> node --experimental-vm-modules node_modules/.bin/jest --coverage --coverageReporters=text-summary

F:\PROJECTS\GSoC\accord\template-engine\node_modules\.bin\jest:2
basedir=$(dirname "$(echo "$0" | sed -e 's,\\,/,g')")
          ^^^^^^^

SyntaxError: missing ) after argument list
```

## What's happening

The test script in `package.json` runs `node node_modules/.bin/jest`. On Linux/macOS this works because `.bin/jest` is a symlink to the actual JS file. On Windows, `.bin/jest` is a bash shell shim, so Node tries to parse it as JavaScript and crashes.

The `coverage` script has the same problem.

Build and lint work fine, only test/coverage are broken.

## Fix

Replace `node_modules/.bin/jest` with `node_modules/jest-cli/bin/jest.js`. That's the actual JS entry point and works on all platforms. The `--experimental-vm-modules` flag stays the same.

I'll submit a PR for this.

## Environment

- Windows 11
- Node v22.16.0
- template-engine v2.7.1

AdityaChauhanX07:
Fixes #129

`node_modules/.bin/jest` is a bash shim on Windows, so `node` can't execute it directly. Changed to `node_modules/jest-cli/bin/jest.js` which is the actual JS entry point and works on all platforms.

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,8 +21,8 @@
     "pretest": "npm-run-all licchk lint build",
     "lint": "eslint .",
     "licchk": "license-check-and-add",
-    "test": "node --experimental-vm-modules node_modules/.bin/jest --coverage --coverageReporters=text-summary",
-    "coverage": "node --experimental-vm-modules node_modules/.bin/jest --coverage",
+    "test": "node --experimental-vm-modules node_modules/jest-cli/bin/jest.js --coverage --coverageReporters=text-summary",
+    "coverage": "node --experimental-vm-modules node_modules/jest-cli/bin/jest.js --coverage",
     "test:watch": "jest --watchAll",
     "updateRuntimeDependencies": "node ./scripts/updateRuntimeDependencies"
   },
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -39,12 +39,12 @@
   "author": "accordproject.org",
   "license": "Apache-2.0",
   "dependencies": {
-    "@accordproject/cicero-core": "0.25.2",
-    "@accordproject/concerto-codegen": "3.32.1",
-    "@accordproject/concerto-core": "3.25.7",
-    "@accordproject/concerto-util":"3.25.7",
-    "@accordproject/markdown-common": "0.17.2",
-    "@accordproject/markdown-template": "0.17.2",
```
mttrbrts, line 47: These changes are unrelated, please revert.

mttrbrts (changes requested):
Jest change looks good, but the dependencies have diverged. Please remove those from the commit before we can merge

mttrbrts (approved):

PR #133, opened 2026-03-31
fix: wrap optional `age` in guard block to fix optional-nested test

### LINKED ISSUE #112: Fix: "optional-nested" test fails after stricter optional checks in PR #53

AbdulRahmanAzam:
## Summary

Fixes the failing `should generate optional-nested` test after PR #53 introduced stricter compile-time checks for optional properties.

## Problem

The `age` property is declared as `optional` in the Concerto model (`model.cto`):
```
@template
concept TemplateData {
    o String name
    o Address address
    o Integer age optional   // <-- optional
    ...
}
```

But in `template.md`, it was used directly without an `{{#optional}}` guard:
```
- You are *{{age}}* years old
```

After PR #53, the template compiler validates that all optional properties are wrapped in `{{#optional}}` blocks, causing this test to fail with:
```
Optional properties used without guards: age
```

## Fix

Wraps the bare `{{age}}` reference in a proper optional guard:
```
- You are *{{#optional age}}{{this}}{{/optional}}* years old
```

This is consistent with the existing usage at the bottom of the same template:
```
{{#optional age}}Age is provided as {{this}}{{else}}Age is hidden{{/optional}}
```

Fixes #112

### DIFF
--- a/test/__snapshots__/TemplateMarkInterpreter.test.ts.snap
+++ b/test/__snapshots__/TemplateMarkInterpreter.test.ts.snap
@@ -3506,47 +3506,42 @@ exports[`templatemark interpreter should generate optional-nested 1`] = `
           ],
         },
         {
-          "$class": "org.accordproject.commonmark@0.5.0.List",
+          "$class": "org.accordproject.commonmark@0.5.0.Paragraph",
           "nodes": [
             {
-              "$class": "org.accordproject.commonmark@0.5.0.Item",
+              "$class": "org.accordproject.ciceromark@0.6.0.Optional",
+              "elementType": "Integer",
+              "hasSome": true,
+              "name": "age",
               "nodes": [
                 {
-                  "$class": "org.accordproject.commonmark@0.5.0.Paragraph",
+                  "$class": "org.accordproject.commonmark@0.5.0.Text",
+                  "text": "- You are ",
+                },
+                {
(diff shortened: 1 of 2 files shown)


mttrbrts (approved):
@AbdulRahmanAzam Thanks for the fix, can you address the DCO sign-off and we're ready to merge.

PR #137, opened 2026-04-15
feat!: upgrade to Concerto v4.0.3 (breaking change)

### LINKED ISSUE #136: Breaking changes for Concerto V4 compatibility and hardening
As discussed in the last work group meeting, I am opening an issue highlighting the API changes I found while testing v4 that need to be made in template engine for it to work with concerto v4. 

First,
ModelManager API update- In v3 we passed the {strict = true} parameter to enforce validation but in v4 the validation strictness was moved out of ModelManager and into the Serializer. So now in v4, the strict=true parameter does not exist inside ModelManager options interface. It breaks the TypeScript build. We need to delete or remove all {strict = true} lines from all the ModelManager instantiations. 

Second,
Serializer API update- As in v4, the Serializer is the new validation gateway, the API has been updated to mandate options object which can be customized by developers. We need to update the Serializer calls in template-engine especially around templateMarkInterpreter to pass an empty options object. This tells v4 to apply the default validation rules. The additional changes in the calls  may look like this:
  `new Serializer(factory, modelManager, {})`
  `serializer.fromJSON(data, {})`

Third, 
Fixing "Optional Guards" Runtime Crashes in Tests- Concerto v4 introduced a strict runtime evaluation rule where if a property is marked optional in a .cto model then template engine will now throw a fatal error if it tries to render the option without checking if it exists first. We need to update the raw .md templates inside template-engine tests which as of now are blindly rendering optional variables without checking if they exist or not. We need to wrap them in a logic block,
(e.g., changing {{middleName}} to {{#if middleName}} {{middleName}} {{/if}}).

But these changes would break template-engine's interactions with the current concerto version that is, v3. So I wanted to ask the maintainers if we need a separate branch of template engine where we can work on making these changes and adjusting it for v4.
@mttrbrts @dselman @DianaLease @muhabulkadir

rockaxorb13:
fixes #136 

This PR fixes the first and second problems listed in #136, as directed by @mttrbrts, these changes did not break the v3 build and as per my understanding, the strict : true parameter for modelManager was mostly redundant.

For the serializer API changes I updated Serializer.fromJSON() calls, adding an option: 
({ validate: true, acceptResourcesForRelationships: false }); 
This satisfies v3's strict TypeScript requirements for the SerializerOptions interface while preparing the calls for v4's updated signature where these options are made optional.

Updated snapshots to adjust these changes.


mttrbrts:
Thanks for working on this! However, adding explicit options like `{ validate: true, acceptResourcesForRelationships: false }` everywhere shouldn't be necessary — these should be the defaults.

The root cause is that the TypeScript type definitions in Concerto v4 appear to incorrectly require these options. I've opened an upstream issue to fix this: https://github.com/accordproject/concerto/issues/1192

Once that's fixed, this PR can be simplified to just remove the `strict: true` from ModelManager (which is the correct fix), without needing to add redundant options to every Serializer call.

Let's fix the type definitions upstream first, then revisit this PR.

rockaxorb13:
Understood! Sorry for the delay, I will get to this asap.

rockaxorb13:
@mttrbrts done, ready for review

mttrbrts (approved):

Issue #135, opened 2026-04-01
Add validation for execution output structure in TemplateArchiveProcessor


Visaka123:
Hi,

While exploring `TemplateArchiveProcessor`, I noticed that the execution result is assumed to always contain `response`, `state`, and `events`, but there is no explicit validation for this structure.

If malformed or incomplete data is returned, it could lead to runtime errors or inconsistent contract state.

### Suggestion
Introduce a validation step after execution to:
- Check for required fields (`response`, `state`, `events`)
- Ensure correct types (e.g., `events` is an array)

### Example
```
ts
const result = await executeLogic(...);
validateExecutionResult(result);
return result;
```
**FIX**

```

function validateExecutionResult(result: any) {
    if (!result || typeof result !== 'object') {
        throw new Error("Invalid execution result");
    }

    if (!('response' in result)) {
        throw new Error("Missing response field");
    }

    if (!('state' in result)) {
        throw new Error("Missing state field");
    }

    if (!Array.isArray(result.events)) {
        throw new Error("Events should be an array");
    }
}

```
his would improve the robustness of the execution pipeline and make it safer for future extensions.

Optional Extension

Such validation could also be useful when integrating external execution mechanisms (e.g., LLM-based reasoning), where output consistency may not always be guaranteed.

Would this be a useful addition? I’d be happy to contribute a PR for this.

Thanks!

mttrbrts:
Fixed by PR #166

PR #144, opened 2026-05-14
Validate template execution results

Rishabh060105:
# Validate TemplateArchiveProcessor execution results

This PR adds the runtime validation requested in #135 so TypeScript template logic fails fast when it returns malformed execution output.

### Changes
- Validates the resolved TypeScript `trigger()` result before returning it:
  - `result` must be an object
  - `state` must be an object
  - `events` must be an array
- Validates that the resolved TypeScript `init()` result contains a state object
- Adds regression coverage for missing `result`, missing `state`, and non-array `events`
- Merges the latest `main` and resolves the previous conflicts

### Related issue
- #135

### Validation
- Targeted processor suite: 15 tests passed
- Full `npm test`: 84 tests and 40 snapshots passed

### Author checklist
- [x] DCO sign-off included
- [x] Tests added for the changed behavior
- [x] Commit messages follow the Accord Project format
- [x] No documentation update required


Rishabh060105:
Hi @mttrbrts , I reviewed the changes against the pointers provided in the original issue and updated the branch accordingly.

### What changed
- Added runtime validation in `TemplateArchiveProcessor.trigger()` so malformed execution output now fails fast instead of being returned unchecked
- Added matching runtime validation in `TemplateArchiveProcessor.init()`
- Covered the new behavior in the real `TemplateArchiveProcessor` test suite for:
  - missing `result`
  - missing `state`
  - non-array `events`

### Validation
- Ran the targeted processor suite:
  - `node --experimental-vm-modules node_modules/jest-cli/bin/jest.js test/TemplateArchiveProcessor.test.ts --runInBand`
- Ran the full repo test command:
  - `npm test`

devanshi00:
Hey @Rishabh060105 , I think this is already covered in pr by @mttrbrts .
You can check this out here: https://github.com/accordproject/template-engine/commit/f5fbfc2082dcfac0b3640c932f49cb162c7f7e02
If I miss something let me know. Otherwise I will close this pr.

Rishabh060105:
Hey @devanshi00,. The PR(PR #166) validates the runtime class hierarchy, while this PR #144  validates the returned execution structure itself.

I’ve resolved all the conflicts, keeping only the missing TypeScript result-shape checks and their tests.It ensures required fields like result, state, and events are present and correctly shaped.

Hi @mttrbrts please review the pr, thanks!

mttrbrts (commented):
I don't think that this PR adds much value, right now, unfortunately.

For the foreseeable future, we expect template logic to be written in Typescript which will mean that bad ouputs are caught by the transpiler. 

Then for the changes specified here, we only check basic properties about the response object. The type checking gives much stronger guarantees.

The author of the original issue, is now protected by the type checking (which wasn't available when they opened the issue). 

Is there a other scenario that your solving for?

Rishabh060105:
Hi @mttrbrts, honestly then I don’t see a strong use case for this PR currently. 

One small gap we could still address is the response envelope where require result, and validate that events is an array only when it is provided. This would make the change more focused without duplicating the current type and hierarchy validation.

mttrbrts:
Thanks for the contribution and for iterating on this, @Rishabh060105 — much appreciated.

Looking into the "does this help plain JavaScript (non-TypeScript) templates" question directly: it doesn't, but not because TS type-checking already covers it — that scenario is actually unreachable today.

`compileLogic()` (`src/TemplateArchiveProcessor.ts`) hard-gates on language:

```ts
if (logicManager.getLanguage() === 'typescript') {
    ...
} else {
    throw new Error('Only TypeScript is supported at this time');
}
```

Both `trigger()` and `init()` call this unconditionally whenever `template.hasLogic()` is true, so a template whose logic is authored in plain JavaScript throws `Only TypeScript is supported at this time` before execution ever reaches `executeTypeScriptTrigger`/`executeTypeScriptInit` — which is where this PR's new `validateTriggerResponse`/`validateInitResponse` checks live. Those checks never get a chance to run for that case; it's rejected upstream for an unrelated reason. (The `JavaScriptEvaluator` used elsewhere only runs compiled output of TypeScript, or compiled TemplateMark formula/condition expressions — never hand-authored JS template logic, consistent with the README's description of templates as TypeScript-only.)

So there isn't a JS-without-TS scenario where this PR's validation adds coverage the compiler doesn't already provide — which lines up with the "not much value right now" read on this PR.

---

PR #149, opened 2026-05-16
fix: draft() emits ciceromark Formula node without required value field (#146)

### LINKED ISSUE #146: draft() emits ciceromark Formula node without required "value" field for inline {{% expr %}} formulas
## Reproduction

Repo: [`accordproject/cicero-template-library`](https://github.com/accordproject/cicero-template-library) PR #484, template `src/supply-agreement-loc`.

`grammar.tem.md` (excerpt):
```
An initial payment of {{% importerLOCAmount / 2.0 %}} will be made immediately upon sight ...
```

`model.cto`:
```
@template
asset TemplateModel extends Clause {
  o Double importerLOCAmount
  ...
}
```

Driver:
```js
const t = await Template.fromDirectory('./src/supply-agreement-loc');
await new TemplateArchiveProcessor(t).draft(data, 'markdown', {});
```

## Observed

```
Error: Generated invalid agreement: ValidationException: The instance "org.accordproject.ciceromark@0.6.0.Formula" is missing the required field "value".: {
  "$class": "org.accordproject.commonmark@0.5.0.Document",
  ...
}
```

The engine produces a `ciceromark@0.6.0.Formula` node and then validates the output; the produced node is missing the required `value` field, so validation fails. The formula has nothing fancy — a single division of a model field by a constant.

## Expected

`draft()` evaluates the inline formula, sets the resulting string on `Formula.value`, and produces a valid AgreementMark document.

## Versions

- `@accordproject/template-engine@3.0.1`
- `@accordproject/cicero-core@0.26.0`
- `@accordproject/concerto-core@4.1.0`

mttrbrts:
Fixes #146.

## Root cause

A formula written as a bare expression — e.g. `{{% importerLOCAmount / 2.0 %}}` without an explicit `return` — was compiled by `writeFunctionToString` into a TS function whose body was an expression statement:

```ts
export function formula_xxx(data, library, options) : any {
    const importerLOCAmount = data.importerLOCAmount;
    // ...
    importerLOCAmount / 2.0   // expression statement — discarded, returns undefined
}
```

That function returns `undefined`. In `evaluateUserCode` the result was then stored as `JSON.stringify(undefined) === undefined`. In the visitor that writes `value` onto the `Formula` node, neither the `result === null` branch nor the `typeof result === 'string'` branch matched, so the fallback `context.value = JSON.stringify(result)` also produced `undefined`. `context.code` was then deleted, leaving the `Formula` node with `name` + `dependencies` but no `value`. `validateCiceroMark` then rejected the document with:

```
ValidationException: The instance "org.accordproject.ciceromark@0.6.0.Formula" is missing the required field "value"
```

This reproduces with the supply-agreement-loc template in cicero-template-library PR #484 (versions: template-engine 3.0.1, cicero-core 0.26.0, concerto-core 4.1.0).

## Fix

- `src/utils.ts`: `writeFunctionToString` now wraps the user's code with `return` when no explicit `return` keyword is present at the source level. Strings and comments are stripped before the keyword scan to avoid false positives (e.g. a literal `'return foo'` inside the user's code). This restores the implicit-expression semantics that Cicero/Ergo formulas have always had — `{{% amount / 2.0 %}}` again produces a value. Existing formulas that already use `return` are unchanged.
- `src/TemplateMarkInterpreter.ts`: as defense in depth, the formula visitor in `generateAgreement` now throws a clear, named error if the formula's evaluation result is still `undefined`, instead of producing an invalid ciceromark document and surfacing the confusing `missing the required field "value"` message from the validator.

## Test

Adds `test/templates/good/helloformula-implicit-return/` — a minimal template containing the inline-formula repro `{{% importerLOCAmount / 2.0 %}}`. The matching snapshot asserts the produced `org.accordproject.ciceromark@0.6.0.Formula` node has `"value": "500"`. Sandbox (child-process) evaluation is also covered by the existing harness, so both code paths are exercised.

The `formula-invalid` snapshot was regenerated: the test still rejects garbage input — only the reported character offsets shift by the length of the prepended `return ` token. No behavior change.

## Scope

Limited to `accordproject/template-engine`. No changes to cicero-core, ciceromark model, concerto, or other dependencies.

## Test plan

- [x] `npx jest TemplateMarkInterpreter -t formula` — all 6 formula tests pass (helloformula, helloformula-implicit-return, formula-now, formula-invalid, formula-no-method, formula-optional-noguard)
- [x] `npm run lint` clean (no new warnings)
- [x] `npm run licchk` clean
- [x] Full `npm test` — remaining 7 failures are pre-existing 403 errors fetching external models (network sandbox) and a pre-existing dynamic-import-callback issue in `TemplateArchiveProcessor.test`; all confirmed unchanged vs. `main` baseline.

---
_Generated by [Claude Code](https://claude.ai/code/session_01TiHU7S7EWmmm3qirCS2Xw6)_


mttrbrts:
Follow-up push: swapped the regex-on-source detection for an AST-based check.

**What changed**
- `wrapExpressionWithReturn` now parses the user code with `ts.createSourceFile(...)` and inspects only `sourceFile.statements` (top-level only).
  - Any top-level `ReturnStatement` → code is kept verbatim.
  - Last top-level statement is an `ExpressionStatement` → emit preceding statements verbatim followed by `return <expr>;`.
  - Otherwise → leave unchanged; the existing "did not return a value" runtime error in `TemplateMarkInterpreter.ts` fires as before.
- Deleted `stripStringsAndComments` and `RETURN_KEYWORD_RE`. `typescript` is already a direct dependency, so no `package.json` changes were needed.

**Why**
The previous regex matched any `return` keyword anywhere in the source (after stripping strings/comments). That triggered a false positive when a nested function or arrow contained its own `return` — e.g. `[1,2,3].map(x => { return x*2 })[0]`. The outer expression was treated as "already returns" and never wrapped, so the resulting Formula node was missing its required `value` field — the same shape of bug this PR was originally fixing, just via a different route. Walking only top-level statements via the TS AST eliminates that class of false positive.

**Tests added** (`test/utils.test.ts`)
- bare expression → wrapped
- explicit top-level `return` → unchanged
- nested arrow with `return`, outer is an expression → wrapped (the regression case)
- multi-statement body with trailing expression → wraps only the trailing expression
- statements-only body (no trailing expression) → unchanged
- empty / whitespace-only → unchanged

**Snapshot update**
The `formula-invalid` snapshot was regenerated. The diff is purely positional for the three semantic `Cannot find name 'THIS'/'IS'/'GARBAGE'` errors, plus the disappearance of a spurious `';' expected.` syntax error — the new wrapping shape (`THIS\nIS\nreturn GARBAGE;`) is itself syntactically valid, so the TS compiler no longer reports a syntax error on top of the identifier errors. No semantic change in the failure mode.

---

PR #154, opened 2026-05-30
fix(test): make external model fixtures hermetic

### LINKED ISSUE #153: Tests are not hermetic: several suites require live model downloads
````md
## Title

Tests are not hermetic: several suites require live model downloads

## Bug Report

### Summary

Some tests in this repository are not hermetic: they require live downloads of external Concerto model files at runtime and fail when network access is unavailable.

I verified this locally by running the affected suites directly in the repository. The failures are reproducible and are caused by remote model fetches rather than local logic regressions.

---

## Reproduction

From the repository root:

```bash
node --experimental-vm-modules node_modules/jest-cli/bin/jest.js --runInBand --coverage=false test/TemplateMarkInterpreter.test.ts

node --experimental-vm-modules node_modules/jest-cli/bin/jest.js --runInBand --coverage=false test/TemplateArchiveProcessor.test.ts
````

---

### `test/TemplateMarkInterpreter.test.ts`

Fails on templates that depend on imported external models.

The failures occur at:

```ts
await modelManager.updateExternalModels();
```

in `test/TemplateMarkInterpreter.test.ts`.

Example error:

```text
Failed to load model file.
Job: https://models.accordproject.org/money@0.3.0.cto
Details: TypeError: fetch failed
```

### `test/TemplateArchiveProcessor.test.ts`

Also fails with remote model fetch errors such as:

```text
Failed to load model file.
Job: https://models.accordproject.org/accordproject/contract@0.2.0.cto
Details: TypeError: fetch failed
```

---

### Known-good suites pass

* `test/HelloWorld.test.ts`
* `test/GenerateOptions.test.ts`

#### `test/TemplateMarkInterpreter.test.ts`

* 24 tests passed
* 7 tests failed

#### `test/TemplateArchiveProcessor.test.ts`

* 1 test passed
* 6 tests failed

This suggests the problem is specifically test hermeticity rather than a general package regression.

---

## Concrete Sources of External Dependency

Examples of fixtures that import remote `.cto` files:

### `test/templates/good/insurance-premium/model.cto`

```cto
import org.accordproject.money@0.3.0.{MonetaryAmount}
from https://models.accordproject.org/money@0.3.0.cto
```

### `test/templates/good/optional-nested/model.cto`

Imports remote:

* money models
* time models
* party models

### `test/archives/latedeliveryandpenalty-typescript/model/model.cto`

Imports remote:

* time models
* contract models
* runtime models

---

## Expected Behavior

The test suite should be runnable without live network access.

Tests should resolve model dependencies from checked-in fixtures, mocked downloads, or another offline-safe mechanism so that CI and local development remain reproducible.

---

## Why This Matters

* Improves CI reliability in sandboxed or restricted environments
* Makes local test runs reproducible
* Avoids false negatives caused by network availability rather than code changes

---

## Possible Fixes

* Vendor the required external `.cto` models into test fixtures
* Remove `updateExternalModels()` from tests where offline fixtures can be used
* Mock or stub external model downloads in tests
* Ensure archive test fixtures fully resolve dependencies offline

---

## Additional Note

Some archive tests already use:

```ts
Template.fromDirectory(..., { offline: true })
```

However, there are still test paths that trigger remote fetches, so offline coverage is currently incomplete.

Rishabh060105:
## Summary

This change makes the test suite hermetic for the remaining external-model paths.

It replaces live model downloads in the affected tests with cached `@models...` fixtures, following the same naming pattern already used in Concerto tests and existing Accord Project archive fixtures.

## Changes

- add cached external Concerto model fixtures under `test/models`
- add a shared test helper to load vendored external models into a `ModelManager`
- mock external model fetches for the archive-based test path using cached local fixtures
- update `TemplateMarkInterpreter.test.ts` to stop calling `updateExternalModels()` at runtime
- update `TemplateArchiveProcessor.test.ts` to resolve external model fetches from local cached fixtures

## Test

```bash
npm test -- --runInBand --coverage=false
```

Results:

- 11 test suites passed
- 63 tests passed
- 38 snapshots passed


Rishabh060105:
Hi @mttrbrts, the fix  follows the Concerto-style cached external model pattern.

What changed:
- vendored the required external models as cached `@models...` fixtures under `test/models`
- replaced the runtime `updateExternalModels()` dependency in `TemplateMarkInterpreter.test.ts` with local cached model loading
- added a shared helper for offline external model resolution in tests
- updated `TemplateArchiveProcessor.test.ts` to resolve external model fetches from cached local fixtures instead of live downloads

I also verified the full suite locally after the change:

```bash
npm test -- --runInBand --coverage=false

**Comment by Rishabh060105** [comment]
@github unstale

**Comment by devanshi00** [comment]
Hey @Rishabh060105, thanks for the PR!

One suggestion before merge: take a look at how cicero-template-library handles this same problem in [test/render.test.mjs](https://github.com/accordproject/cicero-template-library/blob/main/test/render.test.mjs) — it might be a cleaner pattern than fetch-mocking here.

Instead of intercepting fetch and vendoring copies of the external model files, it forces offline: true at the loader boundary: Template.fromDirectory(path, { offline: true }). That routes cicero-core to mm.validateModelFiles() (a local-only check) instead of mm.updateExternalModels() (the network call), so no HTTP request is ever made in the first place. It goes a step further and globally monkeypatches Template.prototype.validate to always pass offline: true, which protects against any internal validate() call elsewhere in the chain, not just the one call site you know about.

Might be worth considering here instead of (or alongside) the mockExternalModelFetches approach , curious what you think.

**Comment by Rishabh060105** [comment]
Hi @devanshi00, `TemplateArchiveProcessor.test.ts` now calls `Template.fromDirectory(..., { offline: true })` and temporarily patches `Template.prototype.validate` to preserve `offline: true` for internal validation calls. The original method is restored afterward to avoid leaking the patch into other tests.

This removes the fetch interception entirely and ensures archive models are validated locally through `validateModelFiles()` without making HTTP requests.

For tests that create a `ModelManager` directly, `offline: true` cannot be used. Instead, the required external model files are loaded locally before the template model. This keeps the tests offline and preserves validation without adding dependencies. Currently, only the money model is needed.

Thank you for pointing it out!
Hi @mttrbrts, please review the pr, thank you!

File: test/TemplateArchiveProcessor.test.ts [hunk 1]
```diff
@@ -1,56 +1,41 @@
...
+        return originalValidate.call(this, {...options, offline: true});
+    };
 
-        return originalFetch(input, init);
-    });
+    try {
+        return await Template.fromDirectory(templateDir, {offline: true});
+    } finally {
+        Template.prototype.validate = originalValidate;
+    }
```
mttrbrts: I worry about the effect of monkey-patching in scenarios where tests are executed in parallel.

Instead, do we understand why the ` Template.fromDirectory(templateDir, {offline: true})` doesn't work as expected?

Rishabh060105: Yes, `Template.fromDirectory(..., { offline: true })` uses offline mode initially, but` cicero-core` later calls `template.validate()` without forwarding the option. `validate()` then falls back to `updateExternalModels()`, causing the network request. 
Basically `cicero-core` is failing to pass  `offline:true` to the `template.validate()` call.We probably need to fix `cicero-core` so the loader forwards offline into its final validation call.

About the monkeypatch: it works well for sequential tests but when tests are executed in parallel it modifies the global state as well.

mttrbrts: Are you referring to this line? https://github.com/accordproject/template-archive/blob/main/packages/cicero-core/src/templateloader.ts#L113

Yes, we should fix that, rather than hack it here.

Rishabh060105: Yes, that is the line. I also found the same issue in the final validation inside `fromDirectory`. I can open a template-archive PR to forward offline through both loader paths and add regression coverage.

devanshi00: Hi @Rishabh060105, please do the required changes here also.
[test/render.test.mjs](https://github.com/accordproject/cicero-template-library/blob/main/test/render.test.mjs)
Thanks!

Rishabh060105:
Hi @mttrbrts, this follow up commit includes updating the `cicero-core` version ,also the removal of global validation monkeypatch.
Also @devanshi00 I'll fix the render.test.mjs in a separate PR.Thank you!!

mttrbrts (approved):
So much better, thanks for doing the right thing here.

PR #155, opened 2026-06-01
feat(logic): decouple compilation for template logic

23harshitkumar:
### Description
This PR addresses the architectural requirements for [US-02](https://github.com/orgs/accordproject/projects/30/views/1?pane=issue&itemId=186684284&issue=accordproject%7Ctemplate-playground%7C898) by strictly decoupling the TS Logic compilation step from the logic execution pipeline (init/trigger methods). 

Previously, `TemplateArchiveProcessor` relied on implicit, iterative compilation hidden inside the `trigger()` and `init()` methods. This PR fully extracts that pipeline into a standalone, manually callable method, allowing upstream platforms (like template-playground) to explicitly validate compilation syntax prior to executing requests.

### Key Changes
- **Decoupled Compilation:** Extracted the internal TypeScript-to-JavaScript compiler loop into a new publicly accessible `compileLogic()` method.
- **Performance Caching:** Introduced a `compiledLogicCache` to preserve compiled `TwoSlashReturn` outputs, eliminating expensive and redundant recompilations during successive `trigger()` calls on the same processor instance.
- **Strict Guard Clauses:** Removed all lazy compilation from `init()` and `trigger()`. Both methods now synchronously throw a descriptive Error if invoked before `compileLogic()` is executed.
- **JSDoc Documentation:** Updated the intellisense WARNING blocks for both init() and trigger() to document the new strictly-enforced prerequisite.
- **Defensive Execution:** Hardened the arguments passed to the `JavaScriptEvaluator` to explicitly resolve fallback values for `currentTime` and `utcOffset`, preventing potential undefined crashes in the V8 isolate.
- **Unit Tests:** Updated all existing test suites to successfully integrate the new `await processor.compileLogic()` requirement, and added new decoupled unit tests explicitly verifying the error rejection handling.

### Context of PR
Guided by @sanketshevkar, this architectural refactor is required for the GSoC Template Playground project. Decoupling the logic compilation step allows the Playground UI to manually compile user-authored TypeScript and surface syntax errors before attempting to execute a mock request. Beyond the frontend benefits, this enforces a strict separation of concerns in the Engine (Build vs. Execute) and drastically improves backend performance by caching compiled logic instead of implicitly recompiling on every trigger.

### DIFF
--- a/src/TemplateArchiveProcessor.ts
+++ b/src/TemplateArchiveProcessor.ts
@@ -44,6 +44,7 @@ export type InitResponse = {
  */
 export class TemplateArchiveProcessor {
     template: Template;
+    private compiledLogicCache?: Record<string, TwoSlashReturn>;
 
     /**
      * Creates a template archive processor
@@ -82,22 +83,22 @@ export class TemplateArchiveProcessor {
 
     }
 
+
     /**
-     * Trigger the logic of a template
-     * @param {object} request - the request to send to the template logic
-     * @param {object} state - the current state of the template
-     * @param {[string]} currentTime - the current time, defaults to now
-     * @param {[number]} utcOffset - the UTC offset, defaults to zero
-     * @returns {Promise} the response and any events
+     * Compile the logic of a template
+     * @returns {Promise<Record<string, TwoSlashReturn>>} the compiled code for each typescript file
      */
-    async trigger(data: any, request: any, state?: any, currentTime?: string, utcOffset?: number): Promise<TriggerResponse> {
+    async compileLogic(): Promise<Record<string, TwoSlashReturn>> {
+        if (this.compiledLogicCache) {
+            return this.compiledLogicCache;
+        }
+
         const logicManager = this.template.getLogicManager();
         if(logicManager.getLanguage() === 'typescript') {
             const compiledCode:Record<string, TwoSlashReturn> = {};
             const tsFiles:Array<Script> = logicManager.getScriptManager().getScriptsForTarget('typescript');
             for(let n=0; n < tsFiles.length; n++) {
                 const tsFile = tsFiles[n];
-                // console.log(`Compiling ${tsFile.getIdentifier()}`);
 
                 const compiler = new TypeScriptToJavaScriptCompiler(this.template.getModelManager(),
                     this.template.getTemplateModel().getFullyQualifiedName());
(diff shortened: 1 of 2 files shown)


23harshitkumar:
@sanketshevkar LMK if this PR meets your expectations or if there are any changes to be made (especially the copilot suggestion regarding lazy compiling and backward compatibility) 

We can continue this discussion here:
https://discord.com/channels/874841541787656263/1511081157314347141

src/TemplateArchiveProcessor.ts
```diff
@@ -106,83 +107,83 @@ export class TemplateArchiveProcessor {
...
-                // console.log(`Compiling ${tsFile.getIdentifier()}`);
-
-                const compiler = new TypeScriptToJavaScriptCompiler(this.template.getModelManager(),
-                    this.template.getTemplateModel().getFullyQualifiedName());
-
-                await compiler.initialize();
+    async trigger(data: any, request: any, state?: any, currentTime?: string, utcOffset?: number): Promise<TriggerResponse> {
+        if (!this.compiledLogicCache) {
+            throw new Error("Logic has not been compiled. You must call compileLogic() before executing init() or trigger().");
+        }
```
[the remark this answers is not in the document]
sanketshevkar: Agree with this, can you please make this change @23harshitkumar

sanketshevkar (commented):
>  Hardened the arguments passed to the JavaScriptEvaluator to explicitly resolve fallback values for currentTime and utcOffset, preventing potential undefined crashes in the V8 isolate.

What's this about, why is this needed? I'm not able to track the change associated with this.

23harshitkumar:
> > Hardened the arguments passed to the JavaScriptEvaluator to explicitly resolve fallback values for currentTime and utcOffset, preventing potential undefined crashes in the V8 isolate.
> 
> What's this about, why is this needed? I'm not able to track the change associated with this.

My apologies! That bullet point was a mistake on my end. While I was rebasing this branch to fix a merge conflict with main, I saw the resolvedTime and resolvedOffset variables that had recently been added to upstream which helped with the case of missing time or timezone offset . I accidentally included that in my PR summary thinking it was part of my local changes. I'll remove that bullet point from the PR description now to avoid any confusion.

23harshitkumar:
**Update: Reverted to Lazy Compilation and Optional `useCache` param**

I've updated the code so we don't accidentally break older apps like apap. I removed the strict error throwing and switched back to lazy compilation, meaning `trigger` and `init` will just automatically handle the compilation if it hasn't been done yet.

To make sure the playground still gets that performance boost, I added an optional `useCache` parameter to those methods. Older apps will ignore it and behave exactly as they did before, but clients that want to opt into the optimization can just pass true to use the cached logic.

src/TemplateArchiveProcessor.ts
```diff
@@ -106,83 +108,79 @@ export class TemplateArchiveProcessor {
...
     /**
-     * Init the logic of a template
+     * Trigger the logic of a template.
+     * @param {object} data - the data for the template
+     * @param {object} request - the request to send to the template logic
+     * @param {object} state - the current state of the template
      * @param {[string]} currentTime - the current time, defaults to now
      * @param {[number]} utcOffset - the UTC offset, defaults to zero
-     * @returns {Promise<InitResponse>} the new state
+     * @param {boolean} [useCache] - whether to use the compiled logic cache
```
sanketshevkar: enableCompiledLogicCache

sanketshevkar (commented):
Nit: Can you please update the argument name across the file and tests? This is the convention we follow to name arguments controlling a behaviour.

src/TemplateArchiveProcessor.ts
23harshitkumar: Sure! I'm pushing the changes.

sanketshevkar (approved):

PR #164, opened 2026-07-10
feat(runtime): validate logic payloads against Concerto models

23harshitkumar:
### Description
This PR addresses the upstream engine requirements for [US-07](https://github.com/orgs/accordproject/projects/30/views/1?pane=issue&itemId=186684458&issue=accordproject%7Ctemplate-playground%7C903): Runtime Payload Validation Against Concerto Models. It ensures that any logic executed by the `TemplateArchiveProcessor` strictly adheres to the underlying Concerto models, both before and after execution.

### Changes Made

- **Input Validation:** Added Concerto Serializer to `init()` and `trigger()`. Payloads (`data`, `request`, `state`) are rigorously validated via `serializer.fromJSON()` before they are passed to the `JavaScriptEvaluator`.

- **Output Validation:** Extracted result, state, and events from the evaluated response and validated them after execution to ensure the logic layer did not mutate them into an invalid schema.

- **Type Checking:** Because we use `serializer.fromJSON()`, this guarantees that the $class discriminators strictly match the declared models in the `ModelManager`.

- **Testing:** Added unit tests to `TemplateArchiveProcessor.test.ts` to verify that invalid payloads (e.g. missing required fields or having wrong data types) throw immediate, descriptive validation errors.

### Note on `cicero-core` typing
I had to add a quick `// @ts-ignore` over the `Script` import in `TemplateArchiveProcessor.ts`. Recent version of `cicero-core` seems to have stopped publicly exporting the internal `types/src/script` path, so TypeScript couldn't find the type definition. I can open a small PR in `cicero-core` to export it from the root index so we can remove this ignore soon. 

### Related Issues
Related to template-playground #[903](https://github.com/orgs/accordproject/projects/30/views/1?pane=issue&itemId=186684458&issue=accordproject%7Ctemplate-playground%7C903)


src/TemplateArchiveProcessor.ts
```diff
@@ -234,23 +236,38 @@ export class TemplateArchiveProcessor {
...
+            triggerResponse = await this.makeLLMExecutor().trigger(data, request, state, currentTime, utcOffset);
+        } else {
+            throw new Error('No executable logic found and LLM fallback is disabled');
         }
 
-        // Otherwise use the LLM executor — either because it was forced, or
-        // because the template carries no executable logic.
-        if (forceLLM || this.shouldUseLLM()) {
-            return this.makeLLMExecutor().trigger(data, request, state, currentTime, utcOffset);
+        // validate outputs after execution
```
mttrbrts, line 263: Do we already have validation inside the LLM Executor? If so, it is safe to assume that the typescript logic will produce valid output. There's an edge case if a template doesn't use Typescript type checking.

I'm happy to be defensive and keep this here, we can come back to optimize later as necessary.

test/TemplateArchiveProcessor.test.ts
```diff
@@ -108,4 +112,46 @@ describe('template archive processor', () => {
...
+            "$identifier": "c88e5ed7-c3e0-4249-a99c-ce9278684ac8"
+        };
+        const invalidRequest = {
+            "$class": "invalid.class.Name", // invalid class
+            "goodsValue": "not a number" // type mismatch
+        };
+
+        const stateResponse = await templateArchiveProcessor.init(validData);
+
+        await expect(templateArchiveProcessor.trigger(validData, invalidRequest, stateResponse.state)).rejects.toThrow();
```
mttrbrts: Check the error message on throw too

mttrbrts (approved):

test/TemplateArchiveProcessor.test.ts
23harshitkumar: Updated using regex. Thanks!

src/TemplateArchiveProcessor.ts
23harshitkumar: No, I checked `LLMExecutor.ts` and it currently doesn't run the raw LLM outputs back through the Concerto `serializer.fromJSON()` for validation. I agree we should keep it here as is.

PR #165, opened 2026-07-12
fix: add strict TemplateLogic compile validation

### LINKED ISSUE #163: Add strict validation for TemplateLogic subclass in TemplateArchiveProcessor compilation
### Description
Currently, when compiling logic via `TemplateArchiveProcessor`, the `template-engine` evaluator supports two flavors of logic: one with a `TemplateLogic` instance and another with a plain JS function. 

However, external execution environments (like the Web Worker Sandbox in `template-playground`) strictly require the compiled output to contain a class extending `TemplateLogic` so it can be instantiated dynamically.

### Proposal
Instead of consumer applications having to manually regex-parse the compiled string (e.g. `code.match(/class\s+(\w+)\s+extends\s+TemplateLogic/)`) to validate the output, it would be much more robust to handle this validation directly within `TemplateArchiveProcessor`.

The compile method could either:
1. Automatically throw a compilation error if a valid `TemplateLogic` class definition isn't found.
2. Accept a strict flag (e.g., `requireTemplateLogic: true`) that enforces this validation during the compilation step.

### Context
This was brought up during the review of the Sandboxed Logic Execution pipeline for the Template Playground: https://github.com/accordproject/template-playground/pull/932#discussion_r3451839005

Rishabh060105:
## Summary

Fixes #163 by adding opt-in strict validation to `TemplateArchiveProcessor.compileLogic()`.

Today `compileLogic()` returns emitted code without checking whether it contains a class extending `TemplateLogic`, so downstream sandbox consumers only discover malformed output later at execution time. This change adds a strict mode for callers that need that guarantee, while keeping the current default behavior unchanged for existing plain-class logic.

## Changes

- add `CompileLogicOptions` to `TemplateArchiveProcessor.compileLogic()`
- keep backward compatibility with the existing boolean cache argument
- support `requireTemplateLogic: true` to enforce that `logic/logic.ts` defines a class extending `TemplateLogic`
- validate the source using the TypeScript AST before returning compiled output
- keep default compilation behavior unchanged so existing plain default-exported classes still compile unless strict mode is enabled
- export `CompileLogicOptions` from the package root
- add regression coverage for:
  - default compile behavior
  - strict compile success when the logic extends `TemplateLogic`
  - strict compile failure for plain default-exported classes

## Test

Implemented and pushed on branch `Rishabh060105/strict-template-logic-validation`.

Validation notes:
- `npm test -- --runTestsByPath test/TemplateArchiveProcessor.test.ts --runInBand`
  - currently blocked by existing upstream build/type issues unrelated to this change
- verified the new `compileLogic()` behavior directly with a focused runtime stub check covering:
  - strict mode success for `TemplateLogic` subclasses
  - default mode success for plain classes
  - strict mode rejection for plain classes
  - cache interaction so strict validation is not bypassed by a previously cached non-strict compile

### DIFF
--- a/src/TemplateArchiveProcessor.ts
+++ b/src/TemplateArchiveProcessor.ts
@@ -27,7 +27,6 @@ import { JavaScriptEvaluator } from './JavaScriptEvaluator';
 import { SMART_LEGAL_CONTRACT_BASE64 } from './runtime/declarations';
 import { LLMExecutor } from './llm/LLMExecutor';
 import { LLMExecutorConfig } from './llm/LLMConfig';
-import * as ts from 'typescript';
 
 /** The contract state. */
 export type State = object;
@@ -80,6 +79,9 @@ export class TemplateArchiveProcessor {
 
     /** Cache of compiled logic, keyed by script identifier. */
     private compiledLogicCache?: Record<string, TwoSlashReturn>;
+
+    /** Tracks whether strict TemplateLogic validation has already passed for this template. */
+    private strictTemplateLogicValidated = false;
 
     /** Optional LLM fallback configuration. */
     llmConfig?: LLMExecutorConfig;
@@ -134,10 +136,11 @@ export class TemplateArchiveProcessor {
         if (logicManager.getLanguage() === 'typescript') {
             const compiledCode: Record<string, TwoSlashReturn> = {};
             const tsFiles: Array<Script> = logicManager.getScriptManager().getScriptsForTarget('typescript');
-            const logicScript = tsFiles.find((tsFile) => tsFile.getIdentifier() === 'logic/logic.ts');
 
-            if (normalizedOptions.requireTemplateLogic) {
-                this.assertTemplateLogicSubclass(logicScript);
+            if (normalizedOptions.requireTemplateLogic && !this.strictTemplateLogicValidated) {
+                const logicScript = tsFiles.find((tsFile) => tsFile.getIdentifier() === 'logic/logic.ts');
+                await this.assertTemplateLogicSubclass(logicScript);
+                this.strictTemplateLogicValidated = true;
             }
 
             if (normalizedOptions.enableCompiledLogicCache && this.compiledLogicCache) {
@@ -168,30 +171,33 @@ export class TemplateArchiveProcessor {
         }
     }
 
-    private assertTemplateLogicSubclass(tsFile?: Script): void {
+    private async assertTemplateLogicSubclass(tsFile?: Script): Promise<void> {
         if (!tsFile) {
             throw new Error('Strict template logic compilation requires a logic/logic.ts file.');
         }
 
-        const sourceFile = ts.createSourceFile(
+        const tsImport = await import('typescript');
+        const tsModule = ('default' in tsImport && tsImport.default ? tsImport.default : tsImport);
+
+        const sourceFile = tsModule.createSourceFile(
             tsFile.getIdentifier(),
             tsFile.getContents(),
-            ts.ScriptTarget.Latest,
+            tsModule.ScriptTarget.Latest,
             true,
-            ts.ScriptKind.TS
+            tsModule.ScriptKind.TS
         );
 
         const hasTemplateLogicSubclass = sourceFile.statements.some((statement) => {
-            if (!ts.isClassDeclaration(statement) || !statement.heritageClauses) {
+            if (!tsModule.isClassDeclaration(statement) || !statement.heritageClauses) {
                 return false;
             }
 
             return statement.heritageClauses.some((clause) =>
-                clause.token === ts.SyntaxKind.ExtendsKeyword &&
+                clause.token === tsModule.SyntaxKind.ExtendsKeyword &&
                 clause.types.some((heritageType) => {
                     const expression = heritageType.expression;
-                    return (ts.isIdentifier(expression) && expression.text === 'TemplateLogic') ||
-                        (ts.isPropertyAccessExpression(expression) && expression.name.text === 'TemplateLogic');
+                    return (tsModule.isIdentifier(expression) && expression.text === 'TemplateLogic') ||
+                        (tsModule.isPropertyAccessExpression(expression) && expression.name.text === 'TemplateLogic');
                 })
             );
         });
--- a/src/TemplateArchiveProcessor.ts
+++ b/src/TemplateArchiveProcessor.ts
@@ -47,28 +47,6 @@ export type InitResponse = {
     state: State;
 }
 
-/** Options for compiling template logic. */
-export type CompileLogicOptions = {
-    enableCompiledLogicCache?: boolean;
-    requireTemplateLogic?: boolean;
-}
-
-type NormalizedCompileLogicOptions = Required<CompileLogicOptions>;
-
-function normalizeCompileLogicOptions(options: boolean | CompileLogicOptions = false): NormalizedCompileLogicOptions {
-    if (typeof options === 'boolean') {
-        return {
-            enableCompiledLogicCache: options,
-            requireTemplateLogic: false
-        };
-    }
-
-    return {
-        enableCompiledLogicCache: options.enableCompiledLogicCache ?? false,
-        requireTemplateLogic: options.requireTemplateLogic ?? false
-    };
-}
-
 /**
  * A template archive processor: can draft content using the
  * templatemark for the archive and trigger the logic of the archive
@@ -79,9 +57,6 @@ export class TemplateArchiveProcessor {
 
     /** Cache of compiled logic, keyed by script identifier. */
     private compiledLogicCache?: Record<string, TwoSlashReturn>;
-
-    /** Tracks whether strict TemplateLogic validation has already passed for this template. */
-    private strictTemplateLogicValidated = false;
 
     /** Optional LLM fallback configuration. */
     llmConfig?: LLMExecutorConfig;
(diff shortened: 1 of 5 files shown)


src/TemplateArchiveProcessor.ts
```diff
@@ -137,6 +168,39 @@ export class TemplateArchiveProcessor {
         }
     }
 
+    private assertTemplateLogicSubclass(tsFile?: Script): void {
+        if (!tsFile) {
+            throw new Error('Strict template logic compilation requires a logic/logic.ts file.');
+        }
+
+        const sourceFile = ts.createSourceFile(
```
23harshitkumar: This looks good! but to optimize bundle size for downstream web consumers (like the playground), could we use a dynamic TS import inside this method instead of the top-level import * as ts?

src/TemplateArchiveProcessor.ts
```diff
@@ -100,19 +123,27 @@ export class TemplateArchiveProcessor {
...
-
+    async compileLogic(options: boolean | CompileLogicOptions = false): Promise<Record<string, TwoSlashReturn>> {
+        const normalizedOptions = normalizeCompileLogicOptions(options);
         const logicManager = this.template.getLogicManager();
         if (logicManager.getLanguage() === 'typescript') {
             const compiledCode: Record<string, TwoSlashReturn> = {};
             const tsFiles: Array<Script> = logicManager.getScriptManager().getScriptsForTarget('typescript');
+            const logicScript = tsFiles.find((tsFile) => tsFile.getIdentifier() === 'logic/logic.ts');
+
+            if (normalizedOptions.requireTemplateLogic) {
```
23harshitkumar: I see you put the validation before the cache check to prevent a previous non-strict compile from bypassing a strict compile later. That is good! but this means that AST will be parsed on every single execution, which defeats the performance gains of the cache.

Could we move the cache check back to the top and just use a boolean flag (like `this.strictValidationPassed = true;`) to track if it's already been validated?

23harshitkumar:
@Rishabh060105 This is great work. Thanks for the contribution! I left some comments!

src/TemplateArchiveProcessor.ts
Rishabh060105: In the follow up commit I removed the top-level `typescript` import and now lazy-load it inside `assertTemplateLogicSubclass()`, so downstream web consumers do not pull it in unless strict validation is actually used.

src/TemplateArchiveProcessor.ts
Rishabh060105: In the follow up commit kept the non-strict -> strict bypass protection, but added a `strictTemplateLogicValidated` flag so cached strict calls do not reparse the AST every time. That lets the cache stay effective after the first strict validation pass.

Rishabh060105:
Hi @23harshitkumar have addressed the issues raised above, please go through it. Thanks!

23harshitkumar:
Thanks @Rishabh060105! You have my approval (as the "issue creator"). 

@sanketshevkar @mttrbrts Could we consider merging it?

mttrbrts:
> Thanks @Rishabh060105! You have my approval (as the "issue creator"). 
> 
> @sanketshevkar @mttrbrts Could we consider merging it? 

I'm tempted to say that this is a bug, and that the opt-in flag is unnecessary. Let's discuss on the WG call tomorrow?

mttrbrts:
> > Thanks @Rishabh060105! You have my approval (as the "issue creator").
> > @sanketshevkar @mttrbrts Could we consider merging it?
> 
> I'm tempted to say that this is a bug, and that the opt-in flag is unnecessary. Let's discuss on the WG call tomorrow?

If these templates will never work without this TemplateLogic extension, then this is a pure bug fix and we don't need the flag to opt-in.

Rishabh060105:
> > > Thanks @Rishabh060105! You have my approval (as the "issue creator").
> > > @sanketshevkar @mttrbrts Could we consider merging it?
> > 
> > 
> > I'm tempted to say that this is a bug, and that the opt-in flag is unnecessary. Let's discuss on the WG call tomorrow?
> 
> If these templates will never work without this TemplateLogic extension, then this is a pure bug fix and we don't need the flag to opt-in.

Hi @mttrbrts , I checked this against both `template-engine` and `template-playground`.

For the playground/sandbox flow, I agree with the concern: that path expects the compiled output to contain a class extending `TemplateLogic`, otherwise it cannot safely derive the class constructor.

For `template-engine` itself though, this is not currently enforced at runtime. `JavaScriptEvaluator` imports the default export, constructs it, and calls the requested method, without checking `extends TemplateLogic`. So a plain default-exported class with the expected methods can still work today.

That is why I kept this as an opt-in flag: it lets sandbox consumers enforce the stricter contract without changing existing engine behavior by default. If we agree with the fact that plain logic classes are unsupported by contract, I’m happy to simplify the PR and make the validation unconditional.

23harshitkumar:
> If these templates will never work without this TemplateLogic extension, then this is a pure bug fix and we don't need the flag to opt-in.

The TemplateLogic extension was enforced in the playground as per the [US:02](https://github.com/orgs/accordproject/projects/30/views/1?pane=issue&itemId=186684284&issue=accordproject%7Ctemplate-playground%7C898) acceptance criteria. This extension helps us in autocomplete and finding out (via regex) which class to instantiate for the sandbox. Therefore, we have an enforcement for this extension in the `store.ts`. So Yes! we do need this extension validation for the playground. 

As for the engine, @Rishabh060105 is right that it currently allows plain classes without failing (could be an accident). To decide if it needs the validation, we'd have to ask ourselves: "do we really want to enforce the TemplateLogic in the engine?". If "yes", then this is a pure bug and we should definately drop the opt-in flag and make the validation unconditional!

mttrbrts:
The evidence suggests that it's accidental that the engine works for classes that don't extend TemplateLogic class. Let's drop the flag. That greatly simplifies the code. Thanks @Rishabh060105!

Rishabh060105:
Hi @mttrbrts, @23harshitkumar  I have pushed the changes:
- Removed the `requireTemplateLogic` opt-in flag.
- `compileLogic()` now always checks that the logic class extends `TemplateLogic`.
- Kept the existing compile cache behavior.

src/worker.js
```diff
@@ -35,12 +35,13 @@ process.on('message', (msg) => {
         const fun = new Function(...argNames, code);
         const result = fun(...args);
         process.send({ result }, () => {
-            process.exit();
+            process.disconnect();
```
mttrbrts: What is the impact of this change?

src/TemplateArchiveProcessor.ts
```diff
@@ -113,6 +113,9 @@ export class TemplateArchiveProcessor {
         if (logicManager.getLanguage() === 'typescript') {
             const compiledCode: Record<string, TwoSlashReturn> = {};
             const tsFiles: Array<Script> = logicManager.getScriptManager().getScriptsForTarget('typescript');
+            const logicScript = tsFiles.find((tsFile) => tsFile.getIdentifier() === 'logic/logic.ts');
+            await this.assertTemplateLogicSubclass(logicScript);
```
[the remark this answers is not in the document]
Rishabh060105: Validation has been kept unconditional as discussed above.The boolean argument now only controls compilation caching.

test/TemplateArchiveProcessor.test.ts
```diff
@@ -57,15 +108,41 @@ describe('template archive processor', () => {
...
+            'assertTemplateLogicSubclass'
+        );
+
+        await templateArchiveProcessor.compileLogic(true);
+        await templateArchiveProcessor.compileLogic(true);
+
+        expect(validationSpy).toHaveBeenCalledTimes(1);
+    });
+
+    test('should reject plain default-exported classes', async () => {
```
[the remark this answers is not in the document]
Rishabh060105: This test is based on the agreed behavior.`compileLogic()` should reject classes that don't extend `TemplateLogic` without the requirement of an opt-in flag.

src/worker.js
```diff
@@ -35,12 +35,13 @@ process.on('message', (msg) => {
...
         });
```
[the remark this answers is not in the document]
Rishabh060105: changed back `process.disconnect()` to `process.exit()` and added a test to confirm the same.

src/worker.js
```diff
@@ -35,12 +35,13 @@ process.on('message', (msg) => {
...
     } catch (err) {
         // console.log(`worker: ${err} ${msg.code}`);
         process.send({ message: err.toString() }, () => {
-            process.exit(1);
+            process.exitCode = 1;
+            process.disconnect();
         });
```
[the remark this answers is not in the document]
Rishabh060105: The error path now uses `process.exit(1)`,and added a test confirming the original error being returned instead of timeout

src/worker.js
Rishabh060105: I initially changed this to `process.disconnect()` because I thought `process.exit()` was interrupting coverage cleanup.I later realised that `process.exit()` is viable here since each worker handles one evaluation this means termination must be deterministic after sending the result.

Rishabh060105:
Hi @mttrbrts, I have addressed the latest review feedback:
- Removed the opt-in flag, `compileLogic()` now always requires a class extending `TemplateLogic`.
- The boolean argument now only controls compilation caching.
- Changed back process.disconnect() to process.exit()
- Added tests covering workers with open timers for both success and error paths.

mttrbrts (approved):

Issue #145, opened 2026-05-15
draft() throws "Paths must be supplied" when {{#ulist}}/{{#olist}} contains direct variable references


mttrbrts:
## Reproduction

Repo: [`accordproject/cicero-template-library`](https://github.com/accordproject/cicero-template-library) PR #484, templates `src/volumediscountulist` and `src/volumediscountolist` (also reproducible against the bootstrapped sample data).

`grammar.tem.md`:
```
{{#ulist rates}}
{{volumeAbove}}\$ million <= Volume < {{volumeUpTo}}\$ million : {{rate}}%
{{/ulist}}
```

`model.cto`:
```
@template
asset TemplateModel extends Clause {
  o RateRange[] rates
}
concept RateRange {
  o Double volumeAbove
  o Double volumeUpTo
  o Double rate
}
```

Minimal driver:
```js
const { Template } = require('@accordproject/cicero-core'); // 0.26.0
const { TemplateArchiveProcessor } = require('@accordproject/template-engine'); // 3.0.1

const t = await Template.fromDirectory('./src/volumediscountulist');
const data = {
  \$class: 'org.accordproject.volumediscountulist@0.1.0.TemplateModel',
  clauseId: 't', \$identifier: 't',
  rates: [
    { \$class: 'org.accordproject.volumediscountulist@0.1.0.RateRange', volumeAbove: 0, volumeUpTo: 1, rate: 3.1 },
    { \$class: 'org.accordproject.volumediscountulist@0.1.0.RateRange', volumeAbove: 1, volumeUpTo: 10, rate: 3.1 },
  ],
};
await new TemplateArchiveProcessor(t).draft(data, 'markdown', {});
```

## Observed

```
Error: Paths must be supplied
    at getJsonPath (.../template-engine/dist/index.js:703:11)
    at Object.<anonymous> (.../template-engine/dist/index.js:889:24)
    at walker (.../traverse/index.js:197:16)
    at Traverse.map (.../traverse/index.js:308:9)
    at generateAgreement (.../template-engine/dist/index.js:819:33)
    at generateRecursiveBlocks (.../template-engine/dist/index.js:799:33)
```

`getJsonPath` (dist line 703) asserts `paths.length >= 1`. Called from the variable-definition branch around line 889 with `this.path` from a `traverse.map`, which is empty when the variable sits directly inside a list block's body.

## Expected

[...]

PR #169, opened 2026-07-26
test(engine): add ordered list regression coverage

coldworld22:
## Summary

Add regression coverage for ordered list rendering when the list body starts with text followed by variables.

## Changes

- Added an ordered-list counterpart to the existing issue #145 regression tests.
- Reused the existing `expectVariablesResolved` helper.
- Verified ordered list generation and variable resolution.

## Testing

- npm test -- TemplateMarkInterpreter

## Notes

This PR only adds test coverage.
No production code, fixtures, or snapshots were changed.

### DIFF
--- a/test/TemplateMarkInterpreter.test.ts
+++ b/test/TemplateMarkInterpreter.test.ts
@@ -141,6 +141,10 @@ concept TemplateData {
                 { $class: 'volumediscount@1.0.0.VolumeDiscount', volumeAbove: 500, rate: 10 },
             ]
         };
+        const LIST_CASES: Array<['ulist' | 'olist', 'bullet' | 'ordered']> = [
+            ['ulist', 'bullet'],
+            ['olist', 'ordered'],
+        ];
 
         async function renderList(content: string): Promise<any> {
             const modelManager = new ModelManager();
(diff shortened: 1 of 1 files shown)


devanshi00:
Hey @coldworld22,
Thanks for adding the relevant test that allows two failure modes get equal test coverage across both list types, rather than leaving one combination (olist + leading text).
Consider parametrizing instead of duplicating. Tests 1/2 and (now) 3/4 are structurally identical pairs that differ only in ulist/olist and 'bullet'/'ordered'. A test.each([['ulist','bullet'], ['olist','ordered']])(...) would remove the duplication and make it structurally impossible for this kind of asymmetry. The next time someone adds a new failure-mode test, they'd add one parametrized case instead of remembering to write both variants. 
It is just a suggestion for future maintainability.
We will still need a review from @mttrbrts before final merge.

coldworld22:
Thanks @devanshi00  for the suggestion. I’ve updated the issue #145 regression tests to use `test.each` for the `ulist`/`olist` cases across both failure modes, so future additions only need one parametrized case.

devanshi00:
Seems perfect for me!
@mttrbrts can you check this out ?

mttrbrts (approved):

PR #173, opened 2026-08-05
feat: Add Template Drafter for `org.accordproject.money@1.0.0.PreciseAmount`

### LINKED ISSUE #172: feat: add drafter for org.accordproject.money@1.0.0.PreciseAmount
## Summary

`org.accordproject.money@1.0.0` introduces `PreciseAmount`, a precise fixed-point monetary type that replaces the `doubleValue`-based `MonetaryAmount` from `0.3.0`. Its Concerto definition carries fields for `value` (string), `scale` (integer), `scheme` (currency code string), etc.

Without a dedicated drafter, templates that bind a `PreciseAmount` variable emit the raw serialised Concerto object — e.g.:

```
{"$class":"org.accordproject.money@1.0.0.PreciseAmount","value":"1234","scale":2,"scheme":"USD"}
```

instead of something human-readable like:

```
1234.00 USD
```

## Proposed solution

Add a `PreciseAmount` drafter in `src/drafting/`, following the same pattern as [`MonetaryAmount`](src/drafting/MonetaryAmount/index.ts), and register it in [`getDrafter()`](src/drafting/index.ts:34):

```typescript
case 'org.accordproject.money@1.0.0.PreciseAmount': return preciseAmountDrafter;
```

### Default rendering (no format string)

Reconstruct the decimal value from `value` and `scale` and append the `scheme`:

```
1234.00 USD
```

i.e. `(BigInt(value) / 10^scale).toFixed(scale) + ' ' + scheme`

### Format string rendering

Reuse the existing `DraftFormat` / `MonetaryAmountFormat` mechanism so template authors can write things like:

```
{{~payment as "K ###,###.00 CCC"~}}
```

producing `$ 1,234.00 USD`.

## Acceptance criteria

- [ ] `PreciseAmount` drafter implemented and registered in `getDrafter()`
- [ ] Default (no-format) rendering reconstructs the decimal correctly from `value`+`scale`
- [ ] Format-string rendering honours `K` (currency symbol) and `CCC` (currency code) tokens, consistent with `MonetaryAmount`
- [ ] Unit tests covering: zero value, positive value, negative value, scale=0, and a custom format string
- [ ] At least one fixture template/data pair in the test corpus demonstrating end-to-end drafting

## References

- Concerto model: https://models.accordproject.org/money@1.0.0.html
- Existing `MonetaryAmount` drafter: `src/drafting/MonetaryAmount/`
- Drafter registry: `src/drafting/index.ts`

ujjwalv01:
Closes #172

## Description
This PR introduces a dedicated template drafter for the new `org.accordproject.money@1.0.0.PreciseAmount` type, allowing templates to render arbitrary-precision monetary amounts as human-readable text.

The `PreciseAmount` model separates the value into an exact `BigInteger` string (`unscaledValue`) and a decimal `scale` (carried on `unit`) to avoid IEEE-754 double-precision drift. This drafter correctly reconstructs the decimal string representation directly from the `unscaledValue` without ever converting to a JavaScript `Number` or `Float` during the default rendering phase, ensuring that precision is fully maintained at any magnitude.

## Key Implementation Details

**1. Arbitrary-Precision String Reconstruction**
A helper function `reconstructDecimalString` was introduced. It uses pure string manipulation to insert the decimal point based on the `unit.scale`, safely handling leading zeroes and negative values without precision loss.

```typescript
function reconstructDecimalString(value: string, scale: number): string {
    if (scale === 0) return value;
    
    let isNegative = false;
    let absValue = value;
    if (absValue.startsWith('-')) {
        isNegative = true;
        absValue = absValue.substring(1);
    }
    
    if (absValue.length <= scale) {
        absValue = absValue.padStart(scale + 1, '0');
    }
    
    const insertPos = absValue.length - scale;
    const integerPart = absValue.substring(0, insertPos);
    const decimalPart = absValue.substring(insertPos);
    
    let result = `${integerPart}.${decimalPart}`;
    return isNegative ? '-' + result : result;
}
```

**2. Formatting Support**
The drafter supports two modes:
* **Default Rendering:** Outputs `reconstructDecimalString(...) + ' ' + unit.code` (e.g. `1234.00 USD`).
* **Format-String Rendering:** For custom formats like `{{payment as "K 0,0.00 CCC"}}`, the drafter re-uses the existing `draftDoubleFormat` utility to replace `K` (currency symbol) and `CCC` (currency code) tokens, ensuring backward compatibility with existing monetary formatting conventions.

**3. Drafter Registration**
The new drafter is registered in `src/drafting/index.ts`:
```typescript
case 'org.accordproject.money@1.0.0.PreciseAmount':
    return preciseAmountDrafter(data, format);
```

## Testing
- **Unit Tests:** Added a comprehensive test suite in `test/PreciseAmountDrafter.test.ts` covering:
  - Zero, positive, and negative `unscaledValue`s
  - Amounts requiring zero-padding (where length of value <= scale)
  - `scale = 0` (no decimal point)
  - Custom format strings
- **End-to-End Fixtures:** Added a new template fixture in `test/templates/good/preciseamount/` containing a `model.cto`, `data.json`, and `template.md` which successfully passed snapshot tests via `TemplateMarkInterpreter`.

## Files Changed
- `src/drafting/PreciseAmount/index.ts` (new) — drafter implementation
- `src/drafting/index.ts` (modified) — drafter registration
- `test/PreciseAmountDrafter.test.ts` (new) — unit tests
- `test/templates/good/preciseamount/` (new) — E2E fixture (`model.cto`, `data.json`, `template.md`)

### DIFF
--- a/src/drafting/MonetaryAmount/formatTokens.ts
+++ b/src/drafting/MonetaryAmount/formatTokens.ts
@@ -1,0 +1,42 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CurrencyCode } from './currencycode';
+
+/**
+ * Symbol from a currency code
+ * @param {string} c - the currency code
+ * @returns {string} the symbol
+ */
+export function codeSymbol(c:string) : string {
+    const index: number = Object.keys(CurrencyCode).indexOf(c);
+    if(index >=0) {
+        return Object.values(CurrencyCode)[index];
+    }
+    else {
+        return c;
+    }
+}
+
+/**
+ * Replaces currency tokens (K for symbol, CCC for code) in a format string
+ * @param {string} format - the format string
+ * @param {string} code - the currency code
+ * @returns {string} the format string with tokens replaced
+ */
+export function replaceCurrencyTokens(format: string, code: string): string {
+    return format
+        .replace(/K/gi, codeSymbol(code))
+        .replace(/CCC/gi, code);
+}
--- a/src/drafting/MonetaryAmount/index.ts
+++ b/src/drafting/MonetaryAmount/index.ts
@@ -15,7 +15,7 @@
 import { draftDoubleIEEE } from '../Double/format';
 import { draftDoubleFormat } from '../Double/format';
 import { MonetaryAmountFormat } from '../DraftFormat';
-import { CurrencyCode } from './currencycode';
+import { replaceCurrencyTokens } from './formatTokens';
 
 type MonetaryAmount = {
     doubleValue: number;
@@ -31,20 +31,6 @@ function monetaryAmountDefaultDrafter(value:MonetaryAmount) {
     return '' + draftDoubleIEEE(value.doubleValue) + ' ' + value.currencyCode;
 }
 
-/**
- * Symbol from a currency code
- * @param {string} c - the currency code
- * @returns {string} the symbol
- */
-function codeSymbol(c:string) : string {
-    const index: number = Object.keys(CurrencyCode).indexOf(c);
-    if(index >=0) {
-        return Object.values(CurrencyCode)[index];
-    }
-    else {
-        return c;
-    }
-}
 
 /**
  * Creates a drafter for monetary amount with a given format
@@ -53,10 +39,7 @@ function codeSymbol(c:string) : string {
  * @returns {string} the text
  */
 function monetaryAmountFormatDrafter(value:MonetaryAmount,format:MonetaryAmountFormat) : string {
-    return draftDoubleFormat(value.doubleValue,
-        format
-            .replace(/K/gi,codeSymbol(value.currencyCode))
-            .replace(/CCC/gi,value.currencyCode));
+    return draftDoubleFormat(value.doubleValue, replaceCurrencyTokens(format, value.currencyCode));
 }
 
 /**
--- a/src/drafting/PreciseAmount/index.ts
+++ b/src/drafting/PreciseAmount/index.ts
@@ -1,0 +1,134 @@
+/*
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { MonetaryAmountFormat } from '../DraftFormat';
+import { replaceCurrencyTokens } from '../MonetaryAmount/formatTokens';
+
+export type PreciseAmount = {
+    unscaledValue: string;
+    unit: {
+        code: string;
+        scheme?: string;
+        identifier?: string;
+        scale: number;
+    };
+};
+
+
+/**
+ * Helper to reconstruct the decimal string representation of a PreciseAmount
+ * without losing precision via JavaScript floats.
+ */
+function reconstructDecimalString(value: string, scale: number): string {
+    if (scale === 0) return value;
+    
+    let isNegative = false;
+    let absValue = value;
+    if (absValue.startsWith('-')) {
+        isNegative = true;
+        absValue = absValue.substring(1);
+    }
+    
+    if (absValue.length <= scale) {
+        absValue = absValue.padStart(scale + 1, '0');
+    }
+    
+    const insertPos = absValue.length - scale;
+    const integerPart = absValue.substring(0, insertPos);
+    const decimalPart = absValue.substring(insertPos);
+    
+    let result = `${integerPart}.${decimalPart}`;
+    if (isNegative) {
+        result = '-' + result;
+    }
+    return result;
+}
+
+/**
+ * Creates a drafter for precise amount with no format
+ * @param {object} value the precise amount
+ * @returns {string} the text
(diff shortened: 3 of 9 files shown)


devanshi00:
Hey @ujjwalv01 , thanks for this pr. I added a few comments.
We will still need a review from @mttrbrts before merging.

ujjwalv01:
Thanks for the detailed Copilot review! Here's a summary of what was addressed:

**Suggestion 1 : - Zero-padding test coverage**
Added the missing test case for the `absValue.length <= scale` branch in `reconstructDecimalString` (e.g. `unscaledValue: "1"`, `scale: 2` → `"0.01 USD"`).

**Suggestion 2  - Missing fixture files**
The fixture files (`model.cto`, `data.json`, `template.md`) under `test/templates/good/preciseamount/` were not committed in the initial push because `test/templates/` is listed in `.gitignore`. They have now been force-added (consistent with how all other existing fixtures in the repo are tracked) and are included in the latest commit.

**Suggestion 3 -  `Number.isFinite` guard for large amounts**
Applied the suggested guard in `preciseAmountFormatDrafter` to prevent a crash when `unscaledValue` is outside the IEEE-754 `Number` range (e.g. very large ETH amounts with 18 decimals). In that case the drafter falls back to the plain `"<value> <code>"` string instead of throwing.

Additionally, two more test cases were proactively added:
- Negative zero-padding: `unscaledValue: "-1"`, `scale: 2` → `"-0.01 USD"`
- `Number.isFinite` fallback: a 400-digit value with a format string correctly returns a plain string instead of `"Infinity USD"`

All 8 unit tests pass.

mttrbrts:
Thanks for iterating on the Copilot feedback — the default-rendering path is solid, and the string-based `reconstructDecimalString` is the right way to preserve precision. A few points remain that I'd like to see addressed before merge, plus one architectural question.

**1. Version dispatch — the drafter is pinned to an exact version (highest priority)**

`getDrafter` matches on the fully-qualified type name including the exact namespace version:

```typescript
case 'org.accordproject.money@1.0.0.PreciseAmount': return preciseAmountDrafter;
```

The moment the money model ships a `1.0.1`, `1.1.0`, or `2.0.0`, `elementType` becomes `org.accordproject.money@1.1.0.PreciseAmount`, which no longer matches. `getDrafter` returns `null` and the interpreter falls back to `JSON.stringify(variableValue)` (`TemplateMarkInterpreter.ts:478`), so the variable renders as a raw JSON blob in the document rather than `1234.00 USD` — silently, with no error. Since `money@1.x` is a new and actively-evolving line, this will bite quickly.

Rather than pinning to `@1.0.0`, I'd suggest normalizing the type name to **namespace + major version** before the switch, so patch/minor releases reuse the drafter and only a breaking major bump (where the shape could legitimately change) requires a new case. `ModelUtil` (already a dependency) has everything needed:

```typescript
import { ModelUtil } from '@accordproject/concerto-core';

function drafterKey(fqn: string): string {
    if (ModelUtil.isPrimitiveType(fqn)) {
        return fqn; // Boolean, String, Integer, ...
    }
    const ns = ModelUtil.getNamespace(fqn);            // org.accordproject.money@1.1.0
    const name = ModelUtil.getShortName(fqn);          // PreciseAmount
    const { name: nsName, version } = ModelUtil.parseNamespace(ns);
    const major = version ? version.split('.')[0] : '';
    return `${nsName}@${major}.${name}`;               // org.accordproject.money@1.PreciseAmount
}
```

The cases then become version-line keys — `org.accordproject.money@1.PreciseAmount`, `org.accordproject.money@0.MonetaryAmount`, etc. This is arguably a pre-existing issue (`MonetaryAmount@0.3.0` is pinned the same way), so it's reasonable to split it into a follow-up — but `PreciseAmount` is the type most exposed to it, so I'd rather not add another exact-version pin. Happy to take it either way as long as we've decided consciously.

**2. Format path still round-trips through `Number` — defeats the type's premise**

`preciseAmountFormatDrafter` does `Number(strValue)` and hands the double to `draftDoubleFormat`. This reintroduces exactly the IEEE-754 precision loss `PreciseAmount` exists to avoid. The `Number.isFinite` guard only catches full overflow to `Infinity` (~309+ digits); the far more common case — any value past ~15–17 significant digits (e.g. an 18-decimal token balance) — rounds *silently* and formats a wrong number that no guard catches. And when the guard does fire, the fallback returns the plain `value + code` string, dropping the requested format entirely. To honor the precision guarantee on the format path too, the grouping/decimal formatting should be applied to the reconstructed *string* (or via `BigInt`) rather than a double. If a full string formatter is out of scope for this PR, that's worth stating explicitly and filing as a follow-up, because as written formatted `PreciseAmount` is no more precise than legacy `MonetaryAmount`.

**3. Duplication with `MonetaryAmount`**

`codeSymbol` is copied verbatim, and the `.replace(/K/gi, …).replace(/CCC/gi, …)` token substitution is duplicated from `monetaryAmountFormatDrafter`. Both should move to a shared helper (e.g. alongside `src/drafting/MonetaryAmount/currencycode.ts`) and be imported by both drafters, so currency-symbol handling has a single source of truth.

**4. Minor:** trailing whitespace inside `reconstructDecimalString` and the trailing blank lines in the test file will likely trip eslint — a local `npm run lint` should catch them.

---

ujjwalv01:
Hi @mttrbrts, thanks for the thorough review! I've addressed all four points in the latest push:

**1. Version Dispatch**
Added a `drafterKey()` helper using `ModelUtil.isPrimitiveType`, `getNamespace`, `getShortName`, and `parseNamespace` — exactly as you suggested. The switch now keys on `org.accordproject.money@1.PreciseAmount` (major-only), so patch and minor releases are handled automatically. I also wrapped it in a `try-catch` to gracefully handle user-defined types without versioned namespaces (which would otherwise throw on `getNamespace`). The existing `MonetaryAmount` and `Duration`/`Period` cases were updated consistently.

**2. Format Path Precision**
Removed the `Number()` conversion entirely from `preciseAmountFormatDrafter`. The formatting is now done directly on the reconstructed decimal string using pure string manipulation — splitting at the decimal point, applying digit grouping, and padding/truncating the decimal part without ever touching a JavaScript float. This preserves full precision even for 18-decimal token balances.

**3. Code Duplication**
Moved `codeSymbol` and the token replacement logic into a new shared file `src/drafting/MonetaryAmount/formatTokens.ts`. Both `MonetaryAmount` and `PreciseAmount` drafters now import `replaceCurrencyTokens` from this single source of truth.

**4. Linting**
Ran `eslint --fix` : trailing whitespace and extra blank lines cleaned up. License headers verified.

All 8 unit tests and all 38 E2E interpreter tests pass. Let me know if there's anything else to adjust!

mttrbrts (approved):
Thanks for the quick turnaround on the feedback. This is looking much stronger now. 

I added some more test cases for some of the more extreme examples too.

PR #174, opened 2026-08-14
[feat]: Add LLM test framework

devanshi00:
### Summary

Adds a Cucumber.js suite that runs Accord Project templates through both the TypeScript executor (`mode: "disabled"`) and the LLM executor (`mode: "force"`), then uses an LLM judge to verify the two outputs are semantically equivalent. Supports both stateless templates (`draft()` + single `trigger()`) and stateful templates (`init()` + an ordered sequence of `trigger()` calls).

### What's included

- **`test/llm_executor/config.js`** — template lookup plus `EXEC_PROVIDERS` / `JUDGE_PROVIDERS`, keyed by provider name (`anthropic`, `google`, `mistral`, `openai`). Models are set via `EXEC_MODEL_<PROVIDER>` / `JUDGE_MODEL_<PROVIDER>` env vars or config defaults — feature files never hardcode a model.
- **`test/llm_executor/judge.js`** — the judge prompt and HTTP call.
- **`test/llm_executor/cucumber.js`** — Cucumber profile, picks up `features/*.feature`.
- **`test/llm_executor/stateless-template_steps.js`** / **`stateful-template_steps.js`** — generic, template-agnostic step definitions.
- **`test/llm_executor/features/`** — two worked examples: `copyright-license.feature` (stateless) and `perishable-goods.feature` (stateful).

### How it works

- Scenarios declare their template by label in the `Given` step (e.g. `Given the stateless template "copyright-license"`), which is joined onto `TEMPLATE_DIR` (defaults to `test/templates`). One run can span many templates by pointing `TEMPLATE_DIR` elsewhere.
- Stateless and stateful step text is intentionally worded differently so both step files can be loaded in a single run without ambiguity.
- For stateful scenarios, each executor initializes and replays the full request sequence against its own independent state; the provider named in `initialize` and `for each request in sequence` steps must match.
- Outputs land under `<template-dir>/responses/disabled/` and `<template-dir>/responses/force/<provider>-output.json`, with non-default request fixtures getting a suffixed copy so runs don't clobber each other.
- Judge verdicts (`equivalent`, `confidence`, `reasoning`, `differences`) are attached to the generated `reports/cucumber-report.html`.

### Running it

```bash
cd test/llm_executor
npm install --save-dev @cucumber/cucumber

# Full run (makes real LLM calls for every scenario)
ANTHROPIC_API_KEY=... npx cucumber-js

# Scoped runs
ANTHROPIC_API_KEY=... npx cucumber-js --tags "@stateless"
ANTHROPIC_API_KEY=... npx cucumber-js --tags "@perishable-goods"
TEMPLATE_DIR=../../examples ANTHROPIC_API_KEY=... npx cucumber-js --tags "@stateful"
```

### Notes for reviewers

- Untagged runs execute every scenario against a real LLM — feature files should be tagged (`@stateless`/`@stateful` + a per-template tag) to keep CI runs scoped and costs predictable.
- `config.js` is the intended single point of edit for adding providers or changing default models/effort settings.
- New templates just need a `*.feature` file under `features/`; no step-definition changes required since the vocabulary is generic.


dselman:
Looks awesome. Will review after we discuss and I see a demo.

DianaLease (approved):
LGTM

DianaLease:
@dselman , I am going to merge this so it isn't blocking @devanshi00 's playground work, but let us know if there is anything you'd like addressed as a fast-follow.

